PR #57, opened 2016-05-04
bring back the curator test.

danyaljj:
fyi @christos-c


mssammon:
when was this removed? Why? (just want to understand implications of bringing it back)

danyaljj:
It used to fail when running semaphore (since curator is not accessible inside semaphore). But then Christos fixed it so that it doesn't work when curator is not available. I don't exactly remember why we kept it commented, but it seems like its working now.

PR #67, opened 2016-05-13
LREC Feature classes

mazinbokhari:
Work in progress


pauljv92:
please note latest commit message

christos-c:
@pauljv92 your package declarations are wrong:
Affixes.java hs `edu.illinois.cs.cogcomp.edison.features.factory.newfexes;`
instead of the correct `edu.illinois.cs.cogcomp.edison.features.lrec;`

Also run the Maven formatter utility (and for the love of god use a modern IDE that won't let you get away with these errors).

pauljv92:
@christos-c  got it, will try to do so to avoid such errors in the future - sometimes eclipse malfunctions when trying to handle large maven projects. The latest commit passes all tests locally, do let me know how I would run the maven formatter utility. 

When I try to run mvn java-formatter:format as per http://code.revelc.net/formatter-maven-plugin/usage.html, I get:

"No plugin found for prefix 'java-formatter' in the current project and in the plugin groups [org.apache.maven.plugins, org.codehaus.mojo] available from the repositories [local (/Users/pauljv/.m2/repository), central (https://repo.maven.apache.org/maven2)] -> [Help 1]"

christos-c:
@pauljv92 the command is `mvn formatter:format`

pauljv92:
@christos-c maven formatting is completed

PR #72, opened 2016-05-19
Merge0513

xinbowu2:
Change implementations of feature extractors to have closer format of features to that of illinois.srl.
  Add a SRL feature comparator to check equality based on contents of features. 
Change all tests to use the SRL feature comparator. 
New tests will not print extracted features, but check features' equality instead.


danyaljj:
@xinbowu2 you have 742 files changed and many of these changes seem to be spurious. Could you close this and open another include *only* your changes?

PR #79, opened 2016-05-21
SRL Implementation in Edison

xinbowu2:
Remove gazetteer and browncluster folders.


xinbowu2:
When running maven install for illinois-core-utility, some errors were produced: 
Failed tests:   testSetDepTreeWithDuplicateTokens(edu.illinois.cs.cogcomp.core.datastructures.textannotation.TreeViewTest): expected:<[Academy]> but was:<[ Academ]>
  testEvaluateNoisySpanLabels(edu.illinois.cs.cogcomp.core.experiments.evaluators.ConstituentLabelingEvaluatorTest): expected:<0.6363636363636364> but was:<0.57>
  testEvaluateSense(edu.illinois.cs.cogcomp.core.experiments.evaluators.PredicateArgumentEvaluatorTest): expected:<0.5> but was:<0.875>

Tests in error: 
  testAddTextAnnotation(edu.illinois.cs.cogcomp.core.io.caches.TextAnnotationDBHandlerTest): org.h2.jdbc.JdbcSQLException: Table "DATASETS" already exists; SQL statement:

I compared my version of these tests and the master and they turned out to be identical. By further observation, the errors seems resulting from an inappropriate implementation of span indexing, which means the way to index a span could not give the result we expected. For example, we actually get " academ", but we need "academy".

mssammon:
I'm going to merge this and then troubleshoot the test failures.  I think there may be an interaction with a change I made a few commits back that affects the behavior of TextAnnotation as it stores span information.

PR #83, opened 2016-05-24
Updated Edison README

christos-c:
Added Maven info, search interface URL, removed spurious technical info
Removed superseded CHANGELOG


edison/README.md
```diff
@@ -15,10 +15,32 @@ make a decision.
 The library is using a Feature Extraction Language (FEX) as a declarative way of defining features. 
 that can be applied to text to generate features. 
 
-## Concepts
-   - Feature extractors
-   - Feature input transformers
-   - Operations
+## Maven dependency
```
danyaljj: Do we want to have these details here? Already mentioned in the root readme file

edison/README.md
```diff
@@ -15,10 +15,32 @@ make a decision.
...
+    <dependency>
+        <groupId>edu.illinois.cs.cogcomp</groupId>
+        <artifactId>illinois-edison</artifactId>
+        <version>3.0.42</version>
+    </dependency>
+</dependencies>
+```
+
+or if you are using SBT:
+```
```
danyaljj: Add this to the root reqme file?

christos-c: I want this information to be here since the poster is pointing people directly to Edison

danyaljj: So maybe change the version number like how its been mentioned in the rootme file, so that we don't have to change it each time?

christos-c: Actually for the time being I want to keep it that way since it makes the copy-paste much easier. In the future we should find a better way to fix this (maybe always point to the latest Maven central released version)

edison/README.md
```diff
@@ -31,48 +53,22 @@ Feature extractors can be combined with each via operators to build newer ones.
...
 ### Bias feature
 The keyword `bias` in a `.fex` specification includes a feature
 that will always be present. This is useful to add a bias feature
 for binary classification.
 
-### Word features
-The following list of feature extractors are operate on the last
-word of the input constituent. They are all defined as static
-members of the class `WordFeatureExtractorFactory`.
-
```
mssammon: I think this table should be preserved somewhere -- these are not part of our new standardized feature extractor implementations.  We can change the javadoc extractor to look in more locations -- right now it only looks under the 'lrec' package -- to find these, but until then...

christos-c: Isn't the idea that everything will be searchable by the form eventually? If so I'd rather have a clean readme now and add this to our search tool later.

mssammon: OK: then can you verify that the tests listed in the table are documented with comparable explanations to those below? if not, let me know and I or someone else can ad them

christos-c: Some of them were there but I've added the rest (in `WordFeatureExtractorFactory`)

PR #93, opened 2016-05-26
Fix: serialization shouldn't fail upon view being null.

danyaljj:
Addendum on #91 
Previous changes in #91 were not quite correct. Fixed the issue and added the case to the UTest. 

@christos-c @mssammon  For whoever has time


mssammon:
Is there a test that verifies the deserialization is correct when a null view is present?

danyaljj:
I added the scenario to the test as well. See the changes.

danyaljj:
Any positive/negative opinions on this?

christos-c:
Can merge but hard to deploy.

PR #97, opened 2016-05-27
More link fixes, and documentation updated.

mayhewsw:
Just changing the NER readme to be more accurate.

### DIFF
--- a/core-utilities/README.md
+++ b/core-utilities/README.md
@@ -83,7 +83,7 @@ of annotations over some text.
     The simplest way to define a `TextAnnotation` is to just give the
     text to the constructor. Note that in the following example,
     `text1` consists of three sentences. The corresponding `ta1` will
-    use the sentence slitter defined in the [Learning based Java](http://cogcomp.cs.illinois.edu/page/software_view/11) (LBJava)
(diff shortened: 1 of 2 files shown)


christos-c:
I think we should move the maven part higher up in the readme file, and also add the repository part of the pom file.

mayhewsw:
I agree about maven. Not sure I understand about "repository part of the pom file"? You mean add the <repository>... cogcomp.../m2repo</repository> thing?

One outstanding problem in this readme is that it is confused about the audience: is it intended for people cloning from github, or downloading from the website?

christos-c:
Yes, I meant adding the `<repositories>` in the pom

I agree that we need to standardise the way we present usage; I think the order should be Maven, then zip download, then source forked/cloned.

danyaljj:
Wow you can assign multiple people to PRs now! 😄

danyaljj:
@christos-c maybe we address ur suggestion in another PR? (and for all the sub-projects)

christos-c:
Ok, but do you mind starting a WIP PR then? I can merge this one after.

danyaljj:
I will create an issue to remember it! 😆

PR #111, opened 2016-06-05
moving  XuePalmber feature extractors to Edison 

danyaljj:
FYI @kordjamshidi @mssammon


kordjamshidi:
@danyaljj , just to remind XuPalmer is not feature extractor, is just a candidate generator.

danyaljj:
Yeah I noticed it; where should it be located?

kordjamshidi:
As I mentioned it is specific for SRL, I am no sure if some other tools will use it.

Issue #112, opened 2016-06-07
Fix NER parameter loading


mayhewsw:
Some parameters are checked incorrectly, for example, "PredictionsLevel1". This can have value 0 or 1, but the code only checks if it is present. If it's not present, the default value is used. This means that it always uses PredictionsLevel1.

File is: Parameters.java

mayhewsw:
Merge #113 fixed this problem.

PR #113, opened 2016-06-07
Fixed parameters.java

mayhewsw:
See #112


cowchipkid:
I tried to look at the mvn output in semaphor to try to identify the failure. The output from the tests is so voluminous, I think it has been truncated.

PR #114, opened 2016-06-07
Fixed H2 DBHelper error

christos-c:
New version of H2 will create a `.mv.db` file instead of the old `.h2.db`

### DIFF
--- a/core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/io/caches/DBHelper.java
+++ b/core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/io/caches/DBHelper.java
@@ -77,7 +77,7 @@ public class DBHelper {
 
     public static boolean dbFileExists(String dbFile) {
         boolean create = false;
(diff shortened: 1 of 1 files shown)


danyaljj:
I'm no expert at this, but seem like [the correct behavior on h2 version](http://stackoverflow.com/a/27100675).

christos-c:
Not sure what you mean. Right now I'm catching both cases, previously only one was caught. I don't want to remove the old one because it's unclear whether it's going to be abandoned in the future.

christos-c:
Btw, testing still fails because of the resources size

danyaljj:
The link says it's either `.h2` **or** `.m2` (while your condition is **and**). As I mentioned I have not used this; not sure how reliable that StackOverFlow link is.... 
Resources: I'm bugging @mssammon about testing [Minio](minio.io) as a datastore, still waiting for engrit ...

christos-c:
I'm doing an & on the negations, which means I'm either failing CS 101 or you are :smile:

danyaljj:
Will merge&deploy tonight.

PR #124, opened 2016-06-29
Added a random tie breaker in ArgMax/Min

christos-c:
Check the unit test for use case.

### DIFF
--- a/core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/math/ArgMax.java
+++ b/core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/math/ArgMax.java
@@ -10,13 +10,16 @@
  */
 package edu.illinois.cs.cogcomp.core.math;
 
+import java.util.Random;
+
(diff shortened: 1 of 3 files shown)


core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/math/ArgMax.java
```diff
@@ -36,34 +39,50 @@
  * </pre>
  *
  * @author Vivek Srikumar
+ * @author Christos Christodouloupoulos
  */
 public class ArgMax<T, K extends Comparable<K>> {
-    private K key;
+    private K value;
```
danyaljj: Rename `K` to `V`, just for ease of understanding?

christos-c: Done

danyaljj:
All good. Merging on green.

PR #131, opened 2016-07-06
ClauseFeatureExtractor now works with full parses

christos-c:
Fixes an issue where the user would not create a PSEUDO_PARSE `View` before extracting features (while the full parse existed).

### DIFF
--- a/edison/src/main/java/edu/illinois/cs/cogcomp/edison/features/factory/ClauseFeatureExtractor.java
+++ b/edison/src/main/java/edu/illinois/cs/cogcomp/edison/features/factory/ClauseFeatureExtractor.java
@@ -46,7 +46,10 @@ public class ClauseFeatureExtractor implements FeatureExtractor {
     public Set<Feature> getFeatures(Constituent c) throws EdisonException {
         TextAnnotation ta = c.getTextAnnotation();
         TreeView pseudoParseView;
-        // If the real parse view exists, use that
+        // If the real parse view exists, use that.
+        // Normally, this feature requires the PSEUDO_PARSE to be manually created by the user
+        // as it is cheaper to make than a full parse. However, there are cases
+        // where the user has already created a full parse view (for other features)
         if (ta.hasView(parseViewName.replace("PSEUDO_", "")))
             pseudoParseView = (TreeView) ta.getView(parseViewName.replace("PSEUDO_", ""));
         else pseudoParseView = (TreeView) ta.getView(parseViewName);
--- a/edison/src/main/java/edu/illinois/cs/cogcomp/edison/features/factory/ClauseFeatureExtractor.java
+++ b/edison/src/main/java/edu/illinois/cs/cogcomp/edison/features/factory/ClauseFeatureExtractor.java
@@ -45,7 +45,11 @@ public class ClauseFeatureExtractor implements FeatureExtractor {
     @Override
     public Set<Feature> getFeatures(Constituent c) throws EdisonException {
         TextAnnotation ta = c.getTextAnnotation();
-        TreeView pseudoParseView = (TreeView) ta.getView(parseViewName);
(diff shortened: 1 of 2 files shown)


edison/src/main/java/edu/illinois/cs/cogcomp/edison/features/factory/ClauseFeatureExtractor.java
```diff
@@ -45,7 +45,11 @@ public ClauseFeatureExtractor(String pseudoParseViewName, String clauseViewName)
     @Override
     public Set<Feature> getFeatures(Constituent c) throws EdisonException {
         TextAnnotation ta = c.getTextAnnotation();
-        TreeView pseudoParseView = (TreeView) ta.getView(parseViewName);
+        TreeView pseudoParseView;
+        // If the real parse view exists, use that
```
danyaljj: could you add a little clarification here why the pseudo-parse might not exist, while the real parse exists?

christos-c: done

edison/src/main/java/edu/illinois/cs/cogcomp/edison/features/factory/ClauseFeatureExtractor.java
```diff
@@ -45,7 +45,14 @@ public ClauseFeatureExtractor(String pseudoParseViewName, String clauseViewName)
     @Override
     public Set<Feature> getFeatures(Constituent c) throws EdisonException {
         TextAnnotation ta = c.getTextAnnotation();
-        TreeView pseudoParseView = (TreeView) ta.getView(parseViewName);
+        TreeView pseudoParseView;
+        // If the real parse view exists, use that.
+        // Normally, this feature requires the PSEUDO_PARSE to be manually created by the user
+        // as it is cheaper to make than a full parse. However, there are cases
+        // where the user has already created a full parse view (for other features)
```
danyaljj: I'm confused now; I thought we build pseudo-parse automatically based on other annotations (e.g. real parse): https://github.com/IllinoisCogComp/illinois-cogcomp-nlp/blob/0965deae01ff6bcd176f96b9fd050369f0ae3e1e/edison/src/main/java/edu/illinois/cs/cogcomp/edison/annotators/PseudoParse.java

christos-c: No, it's based on chunks and POS tags.

danyaljj: ok POS and/or chunks might not be available, in which case we would use the real parse ...

christos-c: Yes, but in general it should be up to the user to decide whether they will add the PSEUDO_PARSE or the real parse view (and their required views) and this feature should be able to use either.

danyaljj: > but in general it should be up to the user to decide whether they will add the PSEUDO_PARSE or the real parse  

Agreed; maybe add a TODO to make it parameterized? (i.e. the target view as parameter)

christos-c: The target view *is* parameterised! It's just that the default constructor use the PSEUDO_PARSE views.

danyaljj: got it; makes sense ...

danyaljj:
Merging on green.

PR #133, opened 2016-07-14
fixes to tokenizer and contraction support

cowchipkid:
Added contraction support, and fixed a few issues with the tokenizer.


mssammon:
This branch fails the integration build because of missing/incorrect license info. You need to run 
mvn license:format 
to automatically update the relevant headers to include the ccg license statement.

cowchipkid:
I am a bit confused about how github works. If I issue a PR then modify and check a file into the branch (or fork in this case) I had requested as the source, it automatically updates the PR to include that modification? Anyway, looks like that’s what happened, so the PR has been updated, and it has not yet completed it’s semaphor build…

danyaljj:
Yup that's how it works

Issue #138, opened 2016-07-21
Location for corpus writers


mayhewsw:
We have a corpusreaders submodule, but I sometimes also want to write to CoNLL NER format from TextAnnotation. Where should this code go? It seems appropriate to put it in the corpusreaders directory, but it's not exactly a reader. 

My vote is to interpret "reader" liberally, as in, "corpus interaction" and include writers in the same directory.

mssammon:
sounds good to me.

PR #139, opened 2016-07-21
Moved files into corpusreaders directory, added conll writer.

mayhewsw:
Moved corpusreaders package out of core-utilities and into corpusreaders.

This caused a problem because of fields in CoNLLColumnFormatReader that were widely referenced. I moved these into PredicateArgumentView, in core-utilities. (The other option was to have core-utilities depend on corpusreaders, but this was a circular dependency).

Also included a TaToConll writer in the CoNLLNerReader file. Fixed the reader to also read sentence boundaries correctly from CoNLL files.


mayhewsw:
This will close #138

danyaljj:
Will merge on green.

mayhewsw:
Just discovered another bug.... hold off on merging for one more commit.

danyaljj:
Do you think the test is general enough to capture these issues? How about making the test more strict? (like instead of just relying on the sizes)

mayhewsw:
In fact, the test would have caught this particular problem... it was an edge case in the data. I changed the data files to reflect it.

mayhewsw:
Build is failing because it can't transfer common-resources to the ci server... any ideas?

danyaljj:
In those case usually a restart fix the CI issue

PR #161, opened 2016-08-19
updated AnnotatorService API, added attributes to Relation

mssammon:
added methods to make AnnotatorService more flexible: get set of views it can provide, allow a new annotator to be added, and allow annotation of an existing TextAnnotation (rather than creating one from scratch or adding views one at a time)
-- this latter is a convenience method and could be removed, provided the getAvailableViews() method is included

### DIFF
--- a/chunker/pom.xml
+++ b/chunker/pom.xml
@@ -2,7 +2,7 @@
     <parent>
         <artifactId>illinois-cogcomp-nlp</artifactId>
         <groupId>edu.illinois.cs.cogcomp</groupId>
-        <version>3.0.61</version>
+        <version>3.0.63-SNAPSHOT</version>
     </parent>
 
     <modelVersion>4.0.0</modelVersion>
@@ -13,7 +13,7 @@
         <dependency>
             <groupId>edu.illinois.cs.cogcomp</groupId>
             <artifactId>illinois-core-utilities</artifactId>
-            <version>3.0.61</version>
+            <version>3.0.63-SNAPSHOT</version>
         </dependency>
 
         <dependency>
@@ -29,7 +29,7 @@
         <dependency>
             <groupId>edu.illinois.cs.cogcomp</groupId>
             <artifactId>illinois-pos</artifactId>
-            <version>3.0.61</version>
+            <version>3.0.63-SNAPSHOT</version>
         </dependency>
         <dependency>
             <groupId>edu.illinois.cs.cogcomp</groupId>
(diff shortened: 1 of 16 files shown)


chunker/pom.xml
```diff
@@ -2,7 +2,7 @@
     <parent>
         <artifactId>illinois-cogcomp-nlp</artifactId>
         <groupId>edu.illinois.cs.cogcomp</groupId>
-        <version>3.0.61</version>
+        <version>3.0.63-SNAPSHOT</version>
```
danyaljj: Version changes are intentional?

mssammon: Yes -- am deploying snapshots so I can work on the project(s) that need(s) these enhancements without delay. I'd prefer not to have to change the version back to the previous one just for PRs.

core-utilities/src/main/java/edu/illinois/cs/cogcomp/annotation/AnnotatorService.java
```diff
@@ -148,4 +149,36 @@ TextAnnotation createAnnotatedTextAnnotation(String corpusId, String textId, Str
...
+     *    built independently of the service, perhaps by a different system component (e.g. a corpus reader).
+     * If so specified, overwrite existing views.
+     *
+     * @param ta The {@link edu.illinois.cs.cogcomp.core.datastructures.textannotation.TextAnnotation} to annotate
+     * @param replaceExistingViews if 'true', annotate a
+     *                              {@link edu.illinois.cs.cogcomp.core.datastructures.textannotation.View} even if
+     *                              it is already present in the ta argument, replacing the original corresponding View.
+     * @return a reference to the updated TextAnnotation
+     */
+    TextAnnotation annotateTextAnnotation(TextAnnotation ta, boolean replaceExistingViews ) throws AnnotatorException;
```
mssammon: boolean replaceExistingViews may be unnecessary -- both Curator and Pipeline have forceUpdate constructor parameter.  But it's not explicit in the API.

chunker/pom.xml
danyaljj: Sounds ok to me; hopefully it won't confusion for other.

danyaljj:
Seems like `CuratorAnnotatorService` needs to be updated too:  
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:2.0.2:compile (default-compile) on project illinois-curator: Compilation failure
[ERROR] /home/runner/illinois-cogcomp-nlp/curator/src/main/java/edu/illinois/cs/cogcomp/curator/CuratorAnnotatorService.java:[34,7] error: CuratorAnnotatorService is not abstract and does not override abstract method annotateTextAnnotation(TextAnnotation,boolean) in AnnotatorService
```

mssammon:
yes, saw that too... fixed now.

cowchipkid:
This looks to be ready to merge, looks good.

danyaljj:
Looks good to me too.

PR #163, opened 2016-08-25
Updated the LBJava dependency to a version that did NOT include the D…

mssammon:
…Vector

which has moved into core. Also change the benchmark code to produce valid
scores at the end of a training run. Also checked in a benchmark.sh script that
works a bit better than the previous one.

NOTE: I'm making this PR on behalf of @cowchipkid


danyaljj:
Looks good to me. Merge upon green.

PR #164, opened 2016-08-26
Adding lbjava tools

danyaljj:
Moving [lbjava-nlp-tools](https://gitlab-beta.engr.illinois.edu/cogcomp/lbjava-nlp-tools) from gitlab. 
No functional change, but transitioning the files. 
Retained the old commits. 

FYI @bhargav


mssammon:
Just want to discuss the issue of circular dependencies here. This change introduces a dependency on lbjava, but the lbjava master project depends on illinois-cogcomp-nlp too.

mssammon:
OK -- ignore my previous comment. Dependency on LBJava is already present though POS etc.

PR #165, opened 2016-08-26
made illinois-common-resources a runtime dependency (trying to avoid …

mssammon:
…the CI downloading this large jar on deploy)

### DIFF
--- a/edison/pom.xml
+++ b/edison/pom.xml
@@ -21,7 +21,7 @@
             <groupId>edu.illinois.cs.cogcomp</groupId>
             <artifactId>illinois-common-resources</artifactId>
             <version>1.3</version>
-            <scope>test</scope>
+            <scope>runtime</scope>
         </dependency>
         <dependency>
             <groupId>edu.illinois.cs.cogcomp</groupId>


cowchipkid:
Good idea, but wondering about the brown cluster data. Gazetteers and models there will be those who do not need them, but I think everybody will need the brown clusters, unless they build their own?

danyaljj:
@cowchipkid we might need either of them in the *runtime*. The annoying thing about current setup is that it download everything in the compile time (e.g. during deploy), which is absurd. Sometimes deploys fail due to issues in downloading such big files.

mssammon:
@cowchipkid as a runtime dependency, the resources should be included in projects that depend on edison, so if anything this change should avoid the problem you identify (previously, this was a 'test'-scoped dependency)

danyaljj:
@mssammon I think this exactly what we want: 
```
runtime is useful for dependencies required for unit tests and at runtime, but not at compile time. This may typically be dynamically loaded code, such as JDBC drivers, which are not directly referenced in the program code.
```

from [here](http://stackoverflow.com/questions/12272499/maven-what-is-the-runtime-scope-purpose).

cowchipkid:
In my case, I use the gaz and the brown clusters, the only thing I don’t use is the models. But I think the gaz is the largest item anyway.

danyaljj:
it fails due to a version bug I introduced unintentionally. Merging this should fix it.

reckart:
How about splitting a large combo-artifact into multiple smaller single-purpose artifacts?

danyaljj:
@reckart I like that, though we'd have soo many small-sized dependencies. @mssammon any comments on that?
Smaller files probably files should have much less change of getting corrupted while downloading....

cowchipkid:
That might help, it might not. The gazetteer is huge, but either way, it's a good idea. I think the original suggestion works for now, maybe as a feature request we split this stuff up down the road.

cowchipkid:
@danyaljj it is more dependencies, however, I wouldn't describe anything in the hundreds of megabytes as small. Question is can they be used independently of each other, and in this case, I think they can. Gaz and brown data will be used in most cases, doubt it will be uncommon for peeps to build their own models though.

mssammon:
There are definitely some distinct resources within illinois-common-resources that could be separated. We would increase the number of dependencies in some projects but it may be worth it. I agree with @cowchipkid that the likely impact is small, though, due to gazetteers being very large even by itself.

reckart:
*Disclaimer: I didn't check out the actual artifacts, so this comment might be redundant.*

JAR files don't compress very well. You could try to compress the gazeteer using [xz (LZMA)](http://tukaani.org/xz/) and store that inside the JAR. xz takes ages for compression but (IMHO) preforms quite decently in terms of the decompression-time/space-saving ratio.

cowchipkid:
@reckart we will need to be able to decompress from within Java. "CORLEX", "lin-clusters" and "lin-similarity", are the largest components in this jar by far. What are they? I don't recall seeing these referenced in NER...

mssammon:
not in NER; some are used in Edison feature extractors and in SRL.

reckart:
There is a [xz Java library on Maven Central](http://search.maven.org/#artifactdetails%7Corg.tukaani%7Cxz%7C1.5%7Cjar). Apache Commons Compress enables its support for xz via that library as well. I've been using xz from Java.

cowchipkid:
@reckart This would require another dependency, I would like to see how much it helped. However, in looking at the contents of that jar, I think this is a non problem. In fact the gazs and brown data are not all that large, but there are other directories of stuff in there, that should be separated out. I vote for compartmentalizing the data, and put only that data that co-occurs in the same jar.

PR #169, opened 2016-08-29
fix lazy init bug in pos tagger, precautionary changes in some other …

### LINKED ISSUE #166: tagger field null even if lazy initialization is disabled
The tagger field in the POSAnnotator remains null after instantiation even if lazy loading is disabled.

Debugging the problem, it seems to turn out that the [field initialization (tagger = null)](https://github.com/IllinoisCogComp/illinois-cogcomp-nlp/commit/6fde90746acba74c7f614ae9d7e39f1c7d79dbbf#diff-0dc695635998a8b8b2a7486515d0c422L39) is executed by the JVM **after** the initialize() method has been invoked (i.e. after the constructor call is complete).

mssammon:
…annotators to avoid potential initialization problems of the same kind. Fixes #166

### DIFF
--- a/chunker/pom.xml
+++ b/chunker/pom.xml
@@ -2,7 +2,7 @@
     <parent>
         <artifactId>illinois-cogcomp-nlp</artifactId>
         <groupId>edu.illinois.cs.cogcomp</groupId>
-        <version>3.0.66</version>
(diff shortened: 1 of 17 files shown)


danyaljj:
All good. Merging.

Issue #160, opened 2016-08-18
Revisit ACE Reader


mssammon:
ACEReader creates a fairly complex TextAnnotation representation over multiple views. It appears that it may make some implicit assumptions about constraints on relations in the source corpus: relation identifiers are stored in argument mentions (as an attribute), which could mean only a single relation per argument can have this information recorded.  Maybe there is a workaround, but I'd like to verify and document it.

bhargav:
Yes. There are a few implementation issues that we should fix to make the Reader simpler:

1. Handling *HEAD* and *EXTENT* for each mention - Right now, we have two different views for mentions - one stores *heads* as constituents and the other one stores *extents* as constituents.  They are both independent views and the only way to reliably map constituents across both views is to use the `EntityMentionID` attribute. 

2. Currently, the `Relation` class that links two constituents does not have support for more attributes. This prevents us from storing *RelationSubType*, *Modality*, *Tense* and other features. Thus, the current workaround is to add new (and potentially duplicate) constituents for each source/target mention. The source mentions is added as a `Predicate` in the Predicate Argument View. Also all relation specific implementation is added to the source mention as attributes.

mssammon:
@bhargav what happens if the same mention is the head of two relations, e.g. "[Professor Roth] works for the University of Illinois and lives in Urbana"?

bhargav:
The PredicateArgumentView would have two different predicates:

- P1 = "[Professor Roth]" predicate with argument as "[University of Illinois]" and the relation specific attributes set inside P1

- P2 = "[Professor Roth]" predicate with argument as "[Urbana]" and the relation specific attributes set inside P2

Since P1 and P2 have different set of attributes, they are treated as different predicates in the PredicateArgumentView. 

Presently, the best way to access all relations would be to get all predicates first and follow their outgoing relations. Each predicate would have a single outgoing relation only. For mentions participating in multiple relations, each relation source would have a separate predicate constituent instance.

bhargav:
Pending Issues with the ACEReader (tracking it here):
- Need to store metadata for each document in the TextAnnotation. [Includes Date/Time, Headline, DocumentID etc]
- The MentionHeadEndCharOffset does not align with character offsets for Token Constituents. Right now the MentionHead attributes are inclusive as [MentionHeadStartCharOffset, MentionHeadEndCharOffset]

mssammon:
I think this is the only remaining issue: 
Pending issues:

[...]

PR #170, opened 2016-08-29
ACEReader fixes 

bhargav:
Fixes towards #160. Simplifies ACE Reader parsing, cleanup unused code and remove some metadata content from **TOKENS** view. @mssammon 

- Replaces `IllinoisTokenizer` with `StatefulTokenizer`
- All Entities and Relations are populated in a single view now. Removes all older views [**This breaks backwords compatibility for ACEReader**]
- Entity Constituents represent EXTENT of the mention. Character Offsets for HEAD is stores as an attribute.
- Relations are represented as `Relation`s in the Entity View directed from "Arg-1" to "Arg-2" role implicitly.
- Entities and Relations have the coarse type as label and fine type is added to attributes.


core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/datastructures/ViewNames.java
```diff
@@ -131,10 +118,7 @@ public static ViewTypes getViewType(String viewName) {
                 return ViewTypes.TOKEN_LABEL_VIEW;
             case SENTENCE:
             case PARAGRAPH:
-            case NER_ACE_COARSE_HEAD:
-            case NER_ACE_FINE_HEAD:
-            case NER_ACE_COARSE_EXTENT:
-            case NER_ACE_FINE_EXTENT:
+            case NER_ACE:
```
mssammon: I think for clarity we should call this "MENTION_ACE" as it has non-Named Entity mentions (correct me if I'm wrong)

mssammon:
@bhargav please see line note about mention view name. 
doc/ACEReader.md needs to be updated to reflect these improvements.
Also: does it make sense to store the metadata fields you now ignore as attribute values in TextAnnotation? (If yes but not urgent, this could be a TODO issue)

core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/datastructures/ViewNames.java
bhargav: Updated the name. It is not strictly NER but also has mentions.

bhargav:
@mssammon Updated the documentation as well.
 
Regarding the ignored attributes, I'll add that as attributes in TextAnnotation. Will send it as a separate PR as it is not urgent right now.

mssammon:
Great -- will merge on green.

PR #171, opened 2016-08-30
fixed bug in ACEReader, added a test that illustrated the problem (wo…

mssammon:
…uld not read hierarchical directories in ACE corpus structure).


corpusreaders/src/main/java/edu/illinois/cs/cogcomp/nlp/corpusreaders/ACEReader.java
```diff
@@ -133,25 +158,27 @@ public boolean accept(File directory, String fileName) {
...
             if (xmlFiles == null) {
                 logger.error("No valid xml file found. Skipping section " + section);
                 continue;
             }
 
-            for (File file : xmlFiles) {
+            for (String fileName : xmlFiles) {
                 ACEDocument doc;
-                String fileName = file.getAbsolutePath();
+//                File file = new File( fileName );
```
bhargav: minor: Remove commented lines here.

bhargav:
Changes look good. @mssammon One question though. We do not have a parser for timex2norm. I wonder how the "apf.xml" file from the test document is being parsed in the test. 

Also thinking if we should remove the `folder.endsWith("bn")` checks inside `ACEFileProcessor` and use some XML field like the `<DOCTYPE SOURCE="newswire"> NEWS STORY </DOCTYPE>` from the .apf.xml files. I'll verify this and add to #160 ticket.

mssammon:
@bhargav maybe this whole change is misguided.  I was working from my copy of the ACE RE project, and there it looks like *only* the timex2norm directories are read for the old ACE system.  I'll take a more careful look and follow up.

PR #182, opened 2016-09-13
replaced / with File.separator

### LINKED ISSUE #31: Fix file path problems in IOUtilss
Same as: https://github.com/cogcomp-dev/illinois-cogcomp-nlp/issues/171 

IOUtils uses some very hacky methods for deciding if something is a directory or a file. This needs to be updated! Line 311 is especially bad. 

Also, all forward slashes need to be replaced with platform agnostic separators.

illinois-cogcomp-nlp/core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/io/IOUtils.java

haowu4:
Fix #31

### DIFF
--- a/core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/io/IOUtils.java
+++ b/core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/io/IOUtils.java
@@ -336,8 +336,8 @@ public abstract class IOUtils {
                 String name = element.getName();
 
                 // FIXME: this is a very bad way to deal with file paths.
-                if (name.startsWith(path) && !name.equals(path + "/")) {
-                    URL url = new URL("jar:" + jarRoot + "/" + name);
+                if (name.startsWith(path) && !name.equals(path + File.separator)) {
+                    URL url = new URL("jar:" + jarRoot + File.separator + name);
                     urls.add(url);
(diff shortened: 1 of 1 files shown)


cogcomp-dev:
@haowu4  please check with @danyaljj to see what else he intended with this ticket (I assume 'hacky' means there are some problems beyond file separators)

haowu4:
Ok, I will talk to him.

haowu4:
Also, many functions in this file is not used at all. I think we should remove them. And add unit test for the rest.
*mkTmpDir
*pwd
*touch
*cleanDir
*rmDir
*lsResourcesDir

What do you think?

danyaljj:
👍 for adding unit test 
not sure abt removing the extra files.

haowu4:
Most of those function are just a fancier version of some instance method of File object.

One more thing I just found is that we should not change "/" to File.separator. Those are Jar entry url, not real file system entry. As a result, they are always seperated by "/".

So we should close or change the description of #31, and ignore this PR.

https://bugs.openjdk.java.net/browse/JDK-6303716

cogcomp-dev:
the functions you list are indeed just decorated versions of File methods, but they nonetheless serve a purpose.  They may not be used in i-c-n, but I certainly use some of them in other projects.

PR #184, opened 2016-09-14
Change the default parameters

yxd126:
The old default value of learning rate and thickness is 0.1 and 20. Now change to 0.05 and 30.

### DIFF
--- a/ner/src/main/lbj/LbjTagger.lbj
+++ b/ner/src/main/lbj/LbjTagger.lbj
@@ -598,6 +598,6 @@ mixed% FeaturesLevel2(NEWord word) <-   PreviousTagPatternLevel2, Level1Aggregat
 discrete NETaggerLevel2(NEWord word)  <-
 learn NELabel 
   using FeaturesLevel1SharedWithLevel2,FeaturesLevel2
-  with new SparseNetworkLearner(new SparseAveragedPerceptron(.1, 0, 20))
+  with new SparseNetworkLearner(new SparseAveragedPerceptron(.05, 0, 30))
 end
 


danyaljj:
@yxd126 do we have enough reason to believe that this change is important/helpful? If so, it's always good to mention them explicitly in the PRs.

yxd126:
@danyaljj Tom has made a table to show different parameters' performance on different data sets. 0.05 and 30 has the highest sum of all the results.  I have attached the table.
[NERParameterSweep.xlsx](https://github.com/IllinoisCogComp/illinois-cogcomp-nlp/files/473292/NERParameterSweep.xlsx)

danyaljj:
@cowchipkid are there any further experiments need to be done for this? 
At least the model should be retrained with these params; right?

mssammon:
@yxd126 did your results come from retraining and evaluating the NER on the relevant data set(s)? Ideally run on all 4 but you should train/evaluate at least CoNLL and OntoNotes.

cowchipkid:
I think this change set will have no effect, since from what I see the java files generated from the lbjava have not been recompiled. Also, the parameters for the level 1 tagger have not been changed.@yxd126 I think I should make these changes since I have already done some of the work, and know what is required. I will make the necessary changes on my fork, and check them in along with the useless ERE readers I have been working on. I will recompile the models and send the results around as well.

mssammon:
@danyaljj @bhargav semaphore indicates build failed due to invalid target release 1.8 -- is this a setting that I need to change?

danyaljj:
Not sure what build status you're talking about ....

PR #191, opened 2016-09-16
ACE Reader fixes

bhargav:
Partly addresses #160 

- Fixed the End Character offset for Entity's HEAD [Attribute]. Made the attribute one-after-the-last-character so that we can treat that as our regular character offset convention.
- Make the ACEReader parse documents lazily and not store the parsed documents in memory. [Helps with reducing memory footprint] : The `hasNext()` method parses one document and keeps a reference to it till the `next()` method is called to retrieve that document.

Pending issues:
- Adding attributes to TextAnnotation and have ACEReader populate document metadata. [Will send a separate PR to simplify review]

### DIFF
--- a/corpusreaders/src/main/java/edu/illinois/cs/cogcomp/nlp/corpusreaders/ACEReader.java
+++ b/corpusreaders/src/main/java/edu/illinois/cs/cogcomp/nlp/corpusreaders/ACEReader.java
@@ -23,8 +23,9 @@ import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.FilenameFilter;
-import java.io.IOException;
 import java.util.*;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 
 /**
  * Corpus reader for the ACE data-set. This reader currently only supports only the ACE-2004 and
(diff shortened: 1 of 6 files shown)


corpusreaders/src/main/java/edu/illinois/cs/cogcomp/nlp/corpusreaders/ACEReader.java
```diff
@@ -74,16 +78,38 @@ public ACEReader(String aceCorpusHome, String[] sections, boolean is2004mode) th
...
         }
 
-        // TODO:Ideally constructor should'nt be reading and processing I/O.
-        this.updateCurrentFiles();
+        File corpusHomeDir = new File(this.aceCorpusHome);
+        assert corpusHomeDir.isDirectory();
+
+        FilenameFilter apfFileFilter = new FilenameFilter() {
+            public boolean accept(File directory, String fileName) {
+                return ( new File(directory + "/" + fileName).isDirectory() || fileName.endsWith(".apf.xml") );
```
danyaljj: `File.Separator`?

danyaljj:
Looks good to me; leaving it to Mark for a final check up.

corpusreaders/src/main/java/edu/illinois/cs/cogcomp/nlp/corpusreaders/ACEReader.java
bhargav: Good catch. Fixed this.

bhargav:
Updated the unit test to check for head string and also added a single ACE2004 document to test resources.

mssammon:
looks good -- do I need to wait for build to pass on Jenkins (as it has apparently passed on semaphore?)

danyaljj:
Ignore Jenkins for now; has some issues stability issues which I have to figure out.

PR #206, opened 2016-09-18
Appveyor CI Integration

bhargav:
- Currently some tests fail but we can have the badge and integration so that each master change is built.


danyaljj:
@bhargav could you make Semaphore happy? Some license header issue; you can probably add sth here: https://github.com/IllinoisCogComp/illinois-cogcomp-nlp/blob/master/pom.xml#L141-L150

Issue #186, opened 2016-09-15
break apart illinois-common-resources


cogcomp-dev:
...into several smaller pieces.  Gazetteers can go in one; brown clusters (and maybe other clusters) in another. This may help reduce the problems with large dependencies in CI.

cowchipkid:
I would suggest anything approaching 100M within this jar file warrants it's own jar file.

mssammon:
There's already a version just for SRL; please check and see if this overlaps with the NER when deciding how to break down the common-resources jar.   The SRL resources jar itself may merit further segmentation.  Before this ticket is closed (after testing NER -- including running the training/evaluation) please open a ticket on the illinois-srl project to migrate to the new jars, if the corresponding resources have been affected.

shatu:
Is there a separate Gitlab project for illinois-common-resources or should I just get it from the repo and then break it apart? And is 1.5 the latest version?

shatu:
BTW, I can also see a illinois-common-resources-1.5-**ner** jar in the repo .. which is around 200M .. any idea what is that being used for?

mssammon:
I think that was me trying to generate an ner-only jar.

shatu:
Okay, so can someone familiar with the following resources, suggest me a logical partitioning, along with the suggested names of the new jars :- 

**Size   	Name (File/Dir)**
 64K   	levin-verbClass.txt (File)
504K   	verb-lemDict.txt (File)
 12M   	cbcData (Dir)

145M   	CORLEX (Dir)

 91M   	WordEmbedding (Dir)

2.6M   	rogetThesaurus (Dir)
 97M   	gazetteers (Dir)

144M   	lin-clusters (Dir)
 57M   	brown-clusters (Dir)

217M   	lin-similarity (Dir)

Also, kindly address the queries posted in the previous comment.

Thanks,
Shashank

cowchipkid:
Definately used the 1.5 version, it includes an updated gazetteer.

I can only speak to a few of resources, but these are the ones I know are used by NER, although there may be others. The brown-clusters and gazetteers are well named and these are used in NER. I would suggest for these two datums, you split them each into their own jar files, I suspect there are cases where each are used independently. I would name them gazetteers.jar and brown-clusters.jar. 

I would suggest you retain the structure within the jar to minimize code changes. In that way, all we should have to change is the pom.xml files…

danyaljj:
[...]

Issue #43, opened 2016-04-27
update common resources


cogcomp-dev:
new version of common resources is available but needs to be deployed. Problem is size: semaphore stops build while downloading it. 
-- should we separate out some different resources to work around this problem?

mssammon:
closed by #186

PR #215, opened 2016-09-20
WIP: Adding SRL

danyaljj:


srl/pom.xml
```diff
@@ -0,0 +1,165 @@
...
+    </properties>
+
+    <dependencies>
+        <!-- Include the pre-trained SRL models for running SemanticRoleLabeler -->
+        <!-- Notice that the models need to match up to the minor version number -->
+        <dependency>
+            <groupId>edu.illinois.cs.cogcomp</groupId>
+            <artifactId>illinois-srl-models</artifactId>
+            <classifier>verb-stanford</classifier>
+            <version>5.1</version>
```
christos-c: Make sure to change the models version (after retraining)

srl/pom.xml
```diff
@@ -0,0 +1,165 @@
...
+        <dependency>
+            <groupId>com.h2database</groupId>
+            <artifactId>h2</artifactId>
+            <version>1.4.190</version>
+        </dependency>
+
+        <dependency>
+            <groupId>edu.illinois.cs.cogcomp</groupId>
+            <artifactId>inference</artifactId>
+            <version>0.5.0</version>
```
christos-c: Is this the latest inference package?

srl/pom.xml
```diff
@@ -0,0 +1,165 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>org.tartarus</groupId>
+            <artifactId>snowball</artifactId>
+            <version>1.0</version>
+        </dependency>
+    </dependencies>
+
+    <build>
```
christos-c: This section is not needed since it's inherited from the parent

srl/pom.xml
```diff
@@ -0,0 +1,165 @@
...
+                    <groupId>edu.illinois.cs.cogcomp</groupId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <!-- The following 3 projects are now developed under illinois-cogcomp-nlp -->
+        <dependency>
+            <groupId>edu.illinois.cs.cogcomp</groupId>
+            <artifactId>illinois-core-utilities</artifactId>
+            <version>${cogcomp-nlp-version}</version>
```
christos-c: These `cogcomp-nlp-version` variables should be replaced with the current super-project version.

srl/pom.xml
```diff
@@ -0,0 +1,165 @@
...
+            <groupId>com.gurobi</groupId>
+            <artifactId>gurobi</artifactId>
+            <version>6.5</version>
+            <optional>true</optional>
+        </dependency>
+
+        <dependency>
+            <groupId>edu.illinois.cs.cogcomp</groupId>
+            <artifactId>illinois-common-resources</artifactId>
+            <classifier>illinoisSRL</classifier>
```
christos-c: Note the use of the classifier in case this needs to be addressed (see #43)

srl/src/main/java/edu/illinois/cs/cogcomp/srl/caches/DBHelper.java
```diff
@@ -0,0 +1,90 @@
+/**
```
christos-c: This exists under `core-utilities` now.

srl/src/main/java/edu/illinois/cs/cogcomp/srl/caches/SentenceDBHandler.java
```diff
@@ -0,0 +1,292 @@
+/**
```
christos-c: This should be moved under `core-utilities`.

srl/src/main/java/edu/illinois/cs/cogcomp/srl/data/FrameData.java
```diff
@@ -0,0 +1,116 @@
+/**
```
christos-c: This should be moved under `corpusreaders`.

srl/src/main/java/edu/illinois/cs/cogcomp/srl/data/FramesManager.java
```diff
@@ -0,0 +1,283 @@
+/**
```
christos-c: This should be moved under `corpusreaders`.

srl/src/main/java/edu/illinois/cs/cogcomp/srl/data/LegalArguments.java
```diff
@@ -0,0 +1,85 @@
...
+
+import java.net.URL;
+import java.util.*;
+
+public class LegalArguments {
+
+	private final static Logger log = LoggerFactory.getLogger(LegalArguments.class);
+
+	private final Map<String, Set<String>> legalArgs = new HashMap<>();
+    private final Map<String, Set<String>> legalArgsForSense = new HashMap<>();
```
christos-c: Indentation error: are we using the `formatter` plugin?

srl/src/main/java/edu/illinois/cs/cogcomp/srl/features/ProjectedPath.java
```diff
@@ -0,0 +1,93 @@
+/**
```
christos-c: This exists in `edison`.

srl/src/main/java/edu/illinois/cs/cogcomp/srl/utilities/WeightVectorUtils.java
```diff
@@ -0,0 +1,134 @@
+/**
```
christos-c: This should be moved to `core-utilities` or `inference` or `illinois-sl`.

srl/src/main/resources/features/Verb.Classifier.old.fex
```diff
@@ -0,0 +1,83 @@
+(define features
```
christos-c: This should be removed.

srl/src/main/resources/features/Verb.Identifier.old.fex
```diff
@@ -0,0 +1,75 @@
+(define features
```
christos-c: This should be removed.

srl/pom.xml
danyaljj: huh? I am not clean on the `classifier` tag usage here ...

srl/src/main/java/edu/illinois/cs/cogcomp/srl/caches/SentenceDBHandler.java
danyaljj: Apparently this is very similar to `TextAnnotationDBHandler.java`. Will try to keep only one of them.

srl/src/main/java/edu/illinois/cs/cogcomp/srl/data/LegalArguments.java
danyaljj: done.

srl/src/main/resources/features/Verb.Identifier.old.fex
danyaljj: done.

srl/src/main/resources/features/Verb.Classifier.old.fex
danyaljj: done.

srl/src/main/java/edu/illinois/cs/cogcomp/srl/utilities/WeightVectorUtils.java
danyaljj: `core-utils` doesn't have `sl` as dependency (and vice versa). Maybe we keep it here for the time being?

srl/pom.xml
christos-c: I was using the `classifier` property to split the common-resources per project. Each project would use its own subunit of the common resources module.

srl/src/main/java/edu/illinois/cs/cogcomp/srl/utilities/WeightVectorUtils.java
christos-c: `illinois-srl` has a dependency to `sl` and `inference` (and `core-utilities` of course). So moving this file to one of those package would be the best solution.

danyaljj: I think my point is different: if you look at the file, it contains functions from both core-utils and SL. Moving this file to one of those projects require one of them to be dependency of the other.

christos-c: oh I see... but the `core-utilities` dependencies are limited to just `IOUtils` and `illinois-sl` already has a solution for these methods (see [WeightVector](https://github.com/IllinoisCogComp/illinois-sl/blob/master/src/main/java/edu/illinois/cs/cogcomp/sl/util/WeightVector.java)). So maybe just a quick fix and this file will be ready to be moved to `sl`

danyaljj: That sounds doable.

srl/pom.xml
danyaljj: screen shot 2016-09-29 at 1 29 46 pm

[1.3 is the one currently used in Edison](https://github.com/IllinoisCogComp/illinois-cogcomp-nlp/blob/master/edison/pom.xml#L20-L25) and [1.4 is the one used in SRL](https://gitlab-beta.engr.illinois.edu/cogcomp/illinois-srl/blob/master/pom.xml#L82-87). There is an effort to break the resources into smaller jar files. [I can update it after that.](https://github.com/IllinoisCogComp/illinois-cogcomp-nlp/issues/186)

christos-c:
@danyaljj I've updated your branch from upstream and updated the cache mechanism.
**NB:** I have NOT tested this on my end since I don't have access to the Propbank/Treebank files; if it doesn't work just let me know and I can try to debug it on my Illinois desktop.

danyaljj:
Yasss! Awesome will test it.

PR #213, opened 2016-09-19
add toTextTable() function

666666fzl:
add toTextTable() method in Table.java, so that it produces text-formatted table

PR #216, opened 2016-09-22
Alternate fix to ParsePhraseTypeOnlyTest

bhargav:
Semaphore test was filing for #213 PR because of an edison unit test failing.

I am not sure if the test always fails but this change should make the test more deterministic.


danyaljj:
Any ideas where the source of non-determinism might be?

danyaljj:
And looks good to me.

Issue #104, opened 2016-05-31
brownBllipClusters removed fron the repo. 


danyaljj:
[brownBllipClusters](https://github.com/IllinoisCogComp/illinois-cogcomp-nlp/blob/master/edison/src/test/resources/brownBllipClusters) are already in the common resources; perhaps we should remove it from the repo (and its history, since its big: 7MB).

PR #219, opened 2016-09-23
Removing brownBllipClusters

cgraber:
Addresses #104 

I used BFG to remove the file from the repo's history; from the website:

> At this point, you're ready for everyone to ditch their old copies of the repo and do fresh clones of the nice, new pristine data. It's best to delete all old clones, as they'll have dirty history that you don't want to risk pushing back into your newly cleaned repo.

So if we're actually serious about removing this from the history people will have to re-clone, as far as I'm aware anyway.


cgraber:
At the behest of @danyaljj , we're not going to remove the file from the commit history. I'll reclone and just delete the file.

PR #220, opened 2016-09-23
Removed brownBllipClusters, since it already exists elsewhere

cgraber:
Addresses #104 . We're not going to remove from the repo history, as that would (potentially) cause many headaches.


danyaljj:
Looks good to me. 
Currently Semaphore is having some issues, unrelated to this PR. We can merge when it becomes happy (perhaps after fixing #186)

PR #119, opened 2016-06-17
Moving big-data-utils 

danyaljj:
FYI @shyamupa @cttsai @nitishgupta @christos-c

PR #227, opened 2016-09-30
Added a TA cache based on MapDB

christos-c:
Created an interface based on `TextAnnotationDBHandler`
The use of the handler is unchanged so far, but it should be theoretically easy to switch to the MapDB-based solution.


mssammon:
As it happens, we have a mapdb-based solution to port from the slot filler validator (thanks to @mpavankumarreddy ) so this is very timely...

danyaljj:
Looks good. Except minor issue: some files miss license header: `mvn license:format`. 

Ideas for future: big-data-utils could be integrated with super-project #119; so instead of adding mapdb directly as a dependency, we can use big-data-utils.

Issue #222, opened 2016-09-29
Fix Chunker evaluation


mssammon:
Fix/streamline testing scripts. Identify the source of variation, and either verify that different performance results are within expected bounds or identify a problem. @qiangning @b29308188  this follows on directly from your recent work on evaluating the Chunker.

b29308188:
@mssammon  [Here](https://github.com/b29308188/illinois-cogcomp-nlp/blob/stable/chunker/src/main/java/edu/illinois/cs/cogcomp/chunker/main/LCTester.java) is my tester.  I have run it several times, but it gave the same results as I wrote in [wiki](https://wiki.illinois.edu/wiki/pages/viewpage.action?spaceKey=ccg&title=Chunker+Benchmark+Results).

mssammon:
@b29308188 @qiangning once you determine the final official eval scripts, please talk with @cowchipkid about correct eval procedure/data splits.

qiangning:
Hi Mark @mssammon, Ben @b29308188 and I just had a brief meeting regarding this problem. I realized that Ben was not aware of this [issue](https://github.com/IllinoisCogComp/illinois-cogcomp-nlp/issues/214). He has fixed this and re-generated his results, which are exactly the same as mine. I have also noted this in our [wiki](https://wiki.illinois.edu/wiki/display/ccg/Chunker+Benchmark+Results). I would suggest a closure for this issue.

danyaljj:
Closed by #233

Issue #221, opened 2016-09-26
illinois-chunker initialization bug


mssammon:
ChunkAnnotator.initialize() uses default constructor, but should take model path from ResourceManager argument.

danyaljj:
Closed by #233

PR #230, opened 2016-10-03
Fixed Chunker Evaluation Iss222 and ChunkAnnotator Init Bug Iss221

qiangning:
It's indeed a git revert thing. I have cleaned things up. Please have a check.
#222 
#221


chunker/scripts/runBenchmarkTest.sh
```diff
@@ -17,7 +17,7 @@ for JAR in `ls $LIB`; do
 done
 
 
-TESTDATA="src/main/resources/test.txt"
+TESTDATA="/shared/corpora/corporaWeb/written/eng/chunking/conll2000distributions/test.txt"
```
danyaljj: this change is intentional?

danyaljj:
I think the tests are failing since your branch `qiangning:master` is way behind `IllinoisCogComp:master`.

chunker/scripts/runBenchmarkTest.sh
qiangning: Yes. But runBenchmarkTest.sh is not used anyhow, you can either adopt it or reject it.

danyaljj: What do you mean by "not used anyhow"? 
Why don't we use only the links to [the files that we already have in the repo](https://github.com/IllinoisCogComp/illinois-cogcomp-nlp/tree/master/chunker/src/test/resources)?

qiangning:
In the beginning I synced my fork with `upstream`, but the resulting package cannot be compiled (so I later reverted the sync). Are you sure that `IllinoisCogComp:master` is good to run? If yes, let me try to redo things tomorrow.

danyaljj:
> Are you sure that `IllinoisCogComp:master` is good to run? 

Do you see the big green badge on the main page of the repo? 

screen shot 2016-10-02 at 8 25 15 pm

> If yes, let me try to redo things tomorrow.

Try a couple of times; if you fail, bring it to me. Git can be weird, but it is important to master. These are the best occasions to learn it.

qiangning:
Thanks for explaining to me. BTW, what do the other two logos saying "build failing" mean? Are they because of my failed pull request?

danyaljj:
Nothing for now. They are not stable, yet.

chunker/scripts/runBenchmarkTest.sh
qiangning: This script is running [ChunkTester](https://github.com/IllinoisCogComp/illinois-cogcomp-nlp/blob/master/chunker/src/main/java/edu/illinois/cs/cogcomp/chunker/main/ChunkTester.java), which doesn't take any file as input. I think one day someone modified the interface of ChunkTester but didn't update runBenchmarkTest.sh.

danyaljj: Ah I see what you saying. Then maybe we can drop `TESTDATA`?

qiangning: Yeah. The reason I'm not worried about it is that I think some day we will add a new ChunkerTester which takes a test file name as input; after all, we want more people using our code.

danyaljj: maybe add a TODO comment to prevent future confusions?

qiangning: Good idea.

PR #234, opened 2016-10-03
New NerBenchmark features and ere-reader

cowchipkid:
Added an ERE reader, but also added a -release command line arg to NerBenchmark to have it build a final release using Test and Train data to train against, and Dev data for automatic convergence.


cowchipkid:
Apparently we are moving to Jenkins? I get a failed check as follows:
ERROR: Couldn't find any revision to build. Verify the repository and branch configuration for this job.
An attempt to send an e-mail to empty list of recipients, ignored.
Finished: FAILURE

holly cow... I was just getting accustomed to semaphore.

danyaljj:
Ignore Jenkins. It's not stable (yet). Only semaphore for now.

danyaljj:
@cowchipkid could you update it with `IllinoisCogcomp:master`?

core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/utilities/TextCleaner.java
```diff
@@ -194,4 +291,24 @@ public static String replaceControlSequence(String origText_) {
         return cleanText;
     }
 
+    
+    /**
+     * Test here.
+     * @param args not used.
+     * @throws Exception 
+     */
+    static public void main (String[] args) throws Exception {
```
danyaljj: why don't we move this into a unit test?

corpusreaders/src/main/java/edu/illinois/cs/cogcomp/nlp/corpusreaders/TokenizationGenerator.java
```diff
@@ -0,0 +1,86 @@
...
+import edu.illinois.cs.cogcomp.nlp.tokenizer.StatefulTokenizer;
+import edu.illinois.cs.cogcomp.nlp.utility.TokenizerTextAnnotationBuilder;
+
+public class TokenizationGenerator {
+    /**
+     * Just for testing.
+     * 
+     * @param args
+     * @throws IOException 
+     */
```
danyaljj: Did you mean to check this in? If so, a little documentation would be very helpful for future.

danyaljj:
Do you think it'd make sense to add a small documentation about this reader's usage? Sth [like this](https://github.com/cowchipkid/illinois-cogcomp-nlp/blob/4d9f744ae51b01e26b6b51f7f5a6717ff094e007/corpusreaders/doc/ACEReader.md).

danyaljj:
Some files are missing license header: `mvn license:format`

cowchipkid:
I believe it is failing because the LBJava generated files don't have license headers in them? How does one fix that?

mssammon:
did you run `mvn license:format`? https://github.com/IllinoisCogComp/illinois-cogcomp-nlp/wiki

cowchipkid:
@mssammon yes I did, but that doesn't do any good for the LbjFeatures files, since these are produced by the LBJava compiler, and (I think), should not be required to have licensing headers...

mssammon:
@cowchipkid so these files are being generated at compile time? It's pretty easy to add the license to java files with `mvn license:format` if they are generated ahead of time and simply deployed.

danyaljj:
@cowchipkid those autogenerated files shouldn't need the license check. Why don't we exclude them? 
https://github.com/IllinoisCogComp/illinois-cogcomp-nlp/blob/master/pom.xml#L141-L152

cowchipkid:
@danyaljj Awesome, I think that is the right way to go.

cowchipkid:
I can not get the "<exclude>" for license headers to work. I have added a few lines, the only one that doesn't work is the one I need to exclude the LBJava generate files in NER from the licensing requirement. Can somebody throw me a bone?
[https://github.com/IllinoisCogComp/illinois-cogcomp-nlp/pull/234/commits/d80cfefd19ef5e66e4e3c4612a92b87bf3e2d884](url)

mssammon:
@mayhewsw and @cttsai please take a look at the ERE reader elements of this PR and give feedback
If you're feeling ambitious you can also look at the NER elements too...

cowchipkid:
And now I am getting an error during "mvn javadoc:test-javadoc", says it can't download a dependency??? This works fine on my machine.

danyaljj:
You can restart the CI in those situations

cowchipkid:
It failed again, same issue. I wonder if it is because we are using the newer LBJava plugin?

danyaljj:
Restarting once or twice often fixed it. Restarted it again ...

danyaljj:
and it's passed. 

@mssammon I think @mayhewsw and @cttsai have a TACL deadline soon. Hopefully they will look at this soon....

ner/src/main/java/edu/illinois/cs/cogcomp/ner/LbjTagger/LearningCurveMultiDataset.java
```diff
@@ -240,21 +239,20 @@ public static void getLearningCurve(Vector<Data> trainDataSet, Vector<Data> test
         }
     }
 
-    /*
+    /**
```
mssammon: If this info isn't already in the NER README, please add it there in a section relating to training -- and give some idea by what you mean by 'a lot of samples'

cowchipkid: The only change I made here was to make the comment a legal javadoc comment. I have no idea what the text actually means. I don't even see a place where the fold sizes are specified in this method. I also turned off the useless output to standard out so we can wade through less text trying to find results. I don't think we are using n-fold methodology with NER at all, we are doing the auto-convergence or specifying the number of iterations.

corpusreaders/src/main/java/edu/illinois/cs/cogcomp/nlp/corpusreaders/TokenizationGenerator.java
cowchipkid: I think I will remove this as well, this was used to evaluate the IllinoisTokenizer against the stageful tokenizer to make a final call on which to use. It could easily be reproduced, but in any event, certainly doesn't belong in this package.

bhargav:
Can't set status; build succeeded.

PR #238, opened 2016-10-11
JSON Serialization Attributes for TextAnnotation, Relation.

bhargav:
- JSON Serialization code was not updated when Attributes were added to
  Relations and TextAnnotations. This PR fixes that and extends the
  existing unit test.


danyaljj:
Looks good. Merging!

Issue #237, opened 2016-10-09
Chunker, NER: method to return set of tags/relation labels that an Annotator can assign


mssammon:
programmatically provide tag set via Annotator class interface. Implement default in Annotator
(empty tag set, warning "not yet implemented") and override in module subclass,
return set of strings corresponding to tags.
Do this for relations also. (will be empty set for many NLP modules).

danyaljj:
@tofuwen has already implemented this for `Annotator`: 
https://github.com/IllinoisCogComp/illinois-cogcomp-nlp/blob/master/core-utilities/src/main/java/edu/illinois/cs/cogcomp/annotation/Annotator.java#L197-L200

The task is to implement this for all the modules that exist in the super-project?

mssammon:
Yes-- the task is to add the functionality for Chunker and NER.  It has already been implemented for POS. @tofuwen , please also add a test for each that verifies that the method is correctly implemented.

danyaljj:
@tofuwen made any progress on this?

tofuwen:
I am working my project for the past week. I will do this today.

tofuwen:
I am almost done for Chunker.

But when I call "doInitialize", there is error:

java.lang.IllegalArgumentException: ERROR: ResourceLoader.getString(): no entry for key 'modelPath'.

	at edu.illinois.cs.cogcomp.core.utilities.configuration.ResourceManager.getString(ResourceManager.java:149)
	at edu.illinois.cs.cogcomp.chunker.main.ChunkerAnnotator.initialize(ChunkerAnnotator.java:66)
	at edu.illinois.cs.cogcomp.annotation.Annotator.doInitialize(Annotator.java:121)
	at edu.illinois.cs.cogcomp.chunker.main.ChunkerAnnotator.getTagValues(ChunkerAnnotator.java:171)
	at edu.illinois.cs.cogcomp.lbj.chunk.tests.TestDiff.testGetTagValues(TestDiff.java:99)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at junit.framework.TestCase.runTest(TestCase.java:168)
[...]

PR #235, opened 2016-10-04
adding get_tag in POSAnnotator

tofuwen:
Closes issue #167

PR #245, opened 2016-10-18
adding tag information for chunker and NERAnnotator

tofuwen:
issue #237

### DIFF
--- a/chunker/src/main/java/edu/illinois/cs/cogcomp/chunker/main/ChunkerAnnotator.java
+++ b/chunker/src/main/java/edu/illinois/cs/cogcomp/chunker/main/ChunkerAnnotator.java
@@ -12,6 +12,7 @@ package edu.illinois.cs.cogcomp.chunker.main;
  */
 
 
+import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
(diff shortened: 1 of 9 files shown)


danyaljj:
Thanks @tofuwen! Why don't we get the tags automatically?

tofuwen:
What do you mean by "automatically"?
The implementation for Chunker and Ner is almost the same for POS.
Do you mean I should add the method in tagger class, just as in POS?

danyaljj:
I'm referring to extracting the tag set, rather than listing them statically in the code.

mssammon:
BTW @danyaljj the listing of tags appears to be in the Test classes, as a way to compare against the automatic retrieval of tags.

danyaljj:
@mssammon @tofuwen Ah I see. My bad ....

tofuwen:
As @bhargav  suggested here: 
https://github.com/IllinoisCogComp/illinois-cogcomp-nlp/pull/235
Which way of implementation is preferred?
I can change all the three implementation according to the suggestion.

@mssammon Are you suggesting that I shouldn't change the constructor, instead of creating a new one?

The issue for the current constructor implementation is that:
If you ever call the default constructor (like "new ChunkerAnnotator()"), it will throw an exception.

And in the test, there is no call to the ChunkerAnnotator. 

Is this a bug, or it's intentionally?

mssammon:
There should be a test for ChunkerAnnotator. 
The default constructor should call another constructor with a default ResourceManager (create from Configurator) and lazyInit flag; this constructor should call super(). This last constructor will use new ChunkerConfigurator().getConfig( nonDefaultConfig) to ensure that all the fields are set wherever it is called from.

mssammon:
Regarding which implementation for getting the tags, from my perspective any of them is fine so long as it has the desired behavior.

bhargav:
@tofuwen In my opinion, you should use either of the two methods that use `.getLabelLexicon()` instead of the `.scores()` on a dummy example. The scores method would need to populate all features for the dummy token/word. 

If the classifier is trained, the LabelLexicon should be populated with all labels that are encountered during training. The .`scores()` method eventually ends up getting all labels from this lexicon and assigning scores corresponding to each label. So directly getting the tagset from the Label Lexicon would be better.

tofuwen:
@bhargav Thanks! I will change it.

danyaljj:
Merge call with Mark.

Issue #183, opened 2016-09-13
Overrride default configurator for POS tagger


mayhewsw:
There is no option to input a config file for the POS tagger. In order for a new user to specify a new training directory, you have to change POSConfigurator and recompile.

b29308188:
@mssammon Should we also modify the main function in order for a new user to specify a new training directory?

mssammon:
main() in which class?

b29308188:
POSTrain

mssammon:
Yes, I think that's a good idea.  It might be best to pass a config file and use it with PosConfigurator supplying default values.  Also, init() should be using constants from POSConfigurator, not string keys, when it gets values from the resourcemanager.

b29308188:
I don't quite understand how to use constant from POSConfigurator.
If we want to use the constructor in ResourceManager : **public ResourceManager(String configFile_)**, then we must provide it with the string name in our configuration file.

For example,
trainingData = /shared/corpora/corporaWeb/written/eng/POS/00-18.br
devData = /shared/corpora/corporaWeb/written/eng/POS/19-21.br
testData = /shared/corpora/corporaWeb/written/eng/POS/22-24.br

may need to take the string names as input.

mssammon:
The POSConfigurator and the keys in the config file must agree. If they don't already, one or the other needs to change. If entries are missing from the POSConfigurator, please add them.

PR #255, opened 2016-10-29
Specify configuration file for POSTrain

b29308188:
issue #183


pos/src/main/java/edu/illinois/cs/cogcomp/pos/POSConfigurator.java
```diff
@@ -66,4 +68,8 @@ public ResourceManager getDefaultConfig() {
                         UNKNOWN_MODEL_PATH, UNKNOWN_LEX_PATH, BASELINE_LEX_PATH, MIKHEEV_LEX_PATH};
         return new ResourceManager(generateProperties(props));
     }
+
```
mssammon: Why is this method needed (i.e. why not just inline "new ResourceManager(configFile)")?

pos/src/main/java/edu/illinois/cs/cogcomp/pos/POSTrain.java
```diff
@@ -36,14 +37,20 @@ public POSTrain() {
 
     public POSTrain(int iter) {
         this.iter = iter;
+        rm = new POSConfigurator().getDefaultConfig();
+        this.init();
+    }
+
+    public POSTrain(int iter, String configFile) throws IOException {
+        this.iter = iter;
+        rm = new POSConfigurator().getConfigFromFile(configFile);
```
mssammon: This line should be "new POSConfigurator().getConfig( new ResourceManager( configFile ) )" -- this way, the user only needs to specify non-default values in the config.

pos/src/main/java/edu/illinois/cs/cogcomp/pos/POSConfigurator.java
b29308188: We may want to have the default configuration even if there's no configuration file path specified in the POStrain.

b29308188:
Modify POSTrain.java and remove getConfigFromFile() method in POSConfigurator.java.

PR #258, opened 2016-11-02
adding cloning to views 

danyaljj:
- overriding clone function for views, relations and constituents 
- using the clone in the evaluators, to prevent them from modifying the contents of the views.

PR #269, opened 2016-11-07
not cleaning view attributes inside evaluators.

danyaljj:
Same as #258

### DIFF
--- a/core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/datastructures/textannotation/Constituent.java
+++ b/core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/datastructures/textannotation/Constituent.java
@@ -179,6 +179,12 @@ public class Constituent implements Serializable, HasAttributes {
         return true;
     }
 
+    /**
+     * This function can be used in scenarios where there is a need for Constituent equality, ignoring the values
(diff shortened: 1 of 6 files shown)


core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/datastructures/textannotation/Constituent.java
```diff
@@ -179,29 +179,7 @@ public boolean doesConstituentCoverAll(Collection<Integer> tokenIds) {
...
-        if (this.attributes == null && that.attributes != null)
-            return false;
-        if (this.attributes != null && that.attributes == null)
-            return false;
-
-        if (this.attributes != null && that.attributes != null)
-            if (!this.attributes.equals(that.attributes))
-                return false;
-
+    public boolean equalsWithoutAttributeEqualityCheck(Constituent that) {
```
bhargav: Can you add some documentation for this? Mentioning the expected use case for using this method.

danyaljj: done.

danyaljj:
Addressed the comment. Planning to merge upon green.

PR #274, opened 2016-11-16
adding label distributions to Constituent and Relation

mssammon:
The label distribution is presently created **on demand** (one less map entry for all cases when only a single label is specified). I.e. either you specify a map of label to score on instantiating the Constituent/Relation and it is explicit from then on, or you only specify a single label, and if someone asks for the label scores a one-entry map is created and stored and returned. 
We *could* make it explicit for single-label instantiation too. @cowchipkid what are your thoughts? 
BTW no idea why all the README changes from @Kevin5025 show up on this PR.


danyaljj:
About Kevin's commits appearing here: I think you didn't checkout the latest version from `IllinoisCogComp:master`.

core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/math/ArgMax.java
```diff
@@ -7,6 +7,10 @@
  */
 package edu.illinois.cs.cogcomp.core.math;
 
+import com.sun.org.apache.xpath.internal.Arg;
```
bhargav: stray import?

core-utilities/README.md
```diff
@@ -346,51 +346,7 @@ for (int i = 0; i < ta.size(); i++) {
 }
 ```
 
-###Creating Annotators
```
danyaljj: @mssammon bring this back?

core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/datastructures/textannotation/Constituent.java
```diff
@@ -45,6 +47,7 @@
 
     protected final String viewName;
     protected Map<String, String> attributes;
+    protected Map<String, Double> labelsToScores;
```
bhargav: For Constituents and Relations, we should also update the serialization to JSON code to save the labelToScore field.

mssammon: good point :-)

core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/datastructures/textannotation/Constituent.java
```diff
@@ -139,6 +159,39 @@ public Constituent(String label, double score, String viewName, TextAnnotation t
                 + ", " + endCharOffset + ")";
     }
 
+    /**
+     * return map of labels to scores. If not explicitly created, generates a trivial distribution
+<<<<<<< HEAD
```
danyaljj: Fix these merge conflict messages

mssammon:
@danyaljj @bhargav I think I responded to everything

bhargav:
Couple of points
- the `hashcode()` methods need to be updated for the new field.
- the `labelToScores` can currently be modified. Should it be an immutable map; to prevent modification? This is assuming that labels (and corresponding) score are assigned when the constituent/relation is created (mostly likely in an output annotator).

danyaljj:
Briefly skimmed through the PR. Looks good to me. 
Merge call with @bhargav

bhargav:
Changes look good. 

@mssammon One small clarification - I see a mapDB-test.db file in the PR but it did not seem to be used anywhere.

PR #281, opened 2016-11-22
minor change to curatoroFactoryTest 

danyaljj:
Putting the tests inside the following block: 
````java 
// if we are running the test on Semaphore, ignore this test, since Gurobi is not provided on Semaphore.
if (System.getenv().containsKey("CI") && System.getenv().get("CI").equals("true")
                 && System.getenv().containsKey("SEMAPHORE")
                && System.getenv().get("SEMAPHORE").equals("true")) {
    System.out.println("Running the test on Semaphore. Skipping this test  . . . ")
} else {
     // test body comes here 
}
````

This will skip tests when running on Semaphore. 
Also FYI [environmental variables on SemaphoreCI](https://semaphoreci.com/docs/available-environment-variables.html).


danyaljj:
FYI, manually checking the a couple of logs, the Semaphore time drops from around ~10min to ~4min :)

PR #283, opened 2016-11-23
make sure TokenizedTextAnnotationBuilder doesn't fail on empty fix.

### LINKED ISSUE #278: Tokenizer breaks on empty strings, carriage returns ...
I experienced that our latest tokenizer breaks on empty string "" and  "\n" at the least.
It produces an 'ArrayOutOfBound Exception'

danyaljj:
Closes #278 

FYI @nitishgupta


danyaljj:
The failure is due to CuratorFactory test, which might be a temporary thing due to connection problem with the Curator. Probably safe to merge, if the changes look fine. (Also Semaphore passes, which skips the Curator tests.)

PR #284, opened 2016-11-24
move pipeline to illinois-cogcomp-nlp

### LINKED ISSUE #168: Pipeline to the superproject
So the bottleneck for current pipeline is that, Semaphore (or any other reasonable CI) is not able to handle such a big memory needed for the tests. 

One idea that came up the other day while talking with @bhargav is that, we can disable such memory intensive tests in the repo, and instead have scripts on our machine running them (on nightly/weekly basis) on a local server. 
So the pipeline will be moved to the superproject, but it will be tested only on our machines once in a while. 

Clearly this is not the safest option, as one might somehow break something in the pipeline; however I think but that should be mostly fine, as we already have test for many of it's components/annotators. 

What are your thoughts on this?

### LINKED ISSUE #241: migrate from h2 to mapdb for BasicAnnotatorService
mapdb appears to have better behavior than h2 regarding scale (can certainly manage 0(10^5) documents without obvious performance bottleneck).  For the record, cannot handle multiple write threads by default.

mssammon:
modified mapdb cache handler
modified BasicAnnotatorService to use mapdb cache
fixed chunker constructors to use configurator properly
moved pipeline to module in master project, updated tests
closes #241 
closes #168


pos/pom.xml
```diff
@@ -26,11 +26,11 @@
             <artifactId>LBJava-NLP-tools</artifactId>
             <version>3.0.84</version>
         </dependency>
-        <dependency>
-            <groupId>edu.illinois.cs.cogcomp</groupId>
-            <artifactId>illinois-pos-models</artifactId>
-            <version>3.0.13</version>
-        </dependency>
+        <!--<dependency>-->
```
danyaljj: I think this is needed?

danyaljj:
One of the tests in `TextAnnotationMapDBHandlerTest ` is failing. It might be due to deletion of  `mapDB-test.db`?  
Also since we are deleting `mapDB-test.db`, maybe we add it to .gitignore file?

pipeline/pom.xml
```diff
@@ -0,0 +1,131 @@
...
+            <classifier>CoNLL</classifier>
+            <version>3.0</version>
+        </dependency>
+
+        <dependency>
+            <groupId>edu.illinois.cs.cogcomp</groupId>
+            <artifactId>illinois-ner-model</artifactId>
+            <version>3.0</version>
+            <classifier>OntoNotes</classifier>
+        </dependency>
```
danyaljj: same as above, dropped?

pipeline/pom.xml
```diff
@@ -0,0 +1,131 @@
...
+
+        <dependency>
+            <groupId>edu.illinois.cs.cogcomp</groupId>
+            <artifactId>illinois-common-resources</artifactId>
+            <version>1.1</version>
+            <scope>runtime</scope>
+        </dependency>
```
danyaljj: I think each module is bringing the own resources?

pipeline/pom.xml
```diff
@@ -0,0 +1,131 @@
...
+            <groupId>com.gurobi</groupId>
+            <artifactId>gurobi</artifactId>
+            <version>6.5</version>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>edu.illinois.cs.cogcomp</groupId>
+            <artifactId>illinois-core-utilities</artifactId>
+            <version>3.0.84</version>
+        </dependency>
```
danyaljj: there is already another "illinois-core-utilities" dependency on top of this file.

pipeline/README.md
```diff
@@ -0,0 +1,301 @@
+# Illinois NLP Pipeline
+
+This module contains code that allows you to run various NLP tools
+to annotate plain text input. 
+
```
danyaljj: Add the link to the root readme file? (Maybe on top of other items?)

pipeline/README.md
```diff
@@ -0,0 +1,301 @@
...
+Gurobi license -- see the section 2.2.)
+
+The download package is organized thus:
+
+config/ : configuration files
+dist/ : the Illinois Preprocessor jar
+lib/ : dependencies 
+scripts/ : scripts to allow command-line test of the Illinois NLP Pipeline
+src/ : source code for the Illinois NLP Pipeline
+test/ : test files used for the command line test of the Illinois NLP Pipeline
```
danyaljj: Add code formatting for these items? Like put in between ```

pipeline/README.md
```diff
@@ -0,0 +1,301 @@
...
+2 Prerequisites
+    2.1  General Requirements: System
+    2.2  Specific Requirements: SRL
+3 Download contents 
+3 Dependencies
+4 Running the Illinois NLP Pipeline
+    4.1  Running a Simple Command-Line Test
+5 Programmatic use
+    5.1 Configuration Options
+    5.2 Changing the logging settings
```
danyaljj: These items are not correctly rendered in the output markdown file. I suspect it's because you have to add an extra space before the numbers.

pipeline/README.md
```diff
@@ -0,0 +1,301 @@
...
+depending on the annotators you plan to use, a reasonable upper limit
+is 1,000 words (fewer if you use resource-intensive annotators like
+Verb or Noun SRL).
+
+The method returns a TextAnnotation data structure (see the 
+illinois-core-utilities package for details), which contains 
+a View corresponding to each annotation source. Each View contains
+a set of Constituents and Relations representing the annotator output.
+Access views and constituents via:
+```
```
danyaljj: add java styling, by adding "java" after ```?

danyaljj:
Thanks Mark! All looks good; just added a few more minor comments to the readme file.

PR #285, opened 2016-11-24
Synchronized BrownCluster init

cowchipkid:
BrownCluster initialization was not sychronized, resulting in race conditions during startup when lazyLoading enabled.

### DIFF
--- a/ner/src/main/java/edu/illinois/cs/cogcomp/ner/ExpressiveFeatures/BrownClusters.java
+++ b/ner/src/main/java/edu/illinois/cs/cogcomp/ner/ExpressiveFeatures/BrownClusters.java
@@ -26,13 +26,18 @@ public class BrownClusters {
     /** the sole instance of this class. */
     private static BrownClusters brownclusters = null;
 
+    /** used to synchronize initialization. */
+    static private final String INIT_SYNC = "Brown Cluster Initialization Synchronization Token";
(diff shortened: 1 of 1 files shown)


cowchipkid:
This did not build correctly on my machine, the POS view was not available when expected during the edu.illinois.cs.cogcomp.curator.CuratorFactoryTest.testGetIndividualTextAnnotationViews. I don't see how this could be the result of the simple change I made, but I am looking into this further.

cowchipkid:
Perhaps it's connectivity issue, mvn build fails on master same as it does on my fork when I run at home.

mssammon:
the Curator POS is running properly, and does not appear to be involved in the CI failure. 
From home, the failure will happen because the test is trying to access the curator across the UIUC/CS firewall.  You should check out your branch on one of our servers and run remotely to verify it's working properly.

This failure looks like a time-out while downloading dependencies -- I'd hoped we were past that particular problem.  I'll force a rebuild and see if the problem remains.

danyaljj:
Looks good to me. One comment befor we merge; is there a way to verify that indeed it fixes the issue? Like maybe a timing experiment?

cowchipkid:
I am reprocessing our repository of 100M documents, massively multi-threaded. Works there. That's how I have found all these race conditions.

PR #293, opened 2016-11-27
NER for cross-lingual wikifier

cttsai:
Modified ner package to support cross-lingual wikifier


ner/src/main/java/edu/illinois/cs/cogcomp/ner/ExpressiveFeatures/GazetteerTree.java
```diff
@@ -35,7 +36,16 @@
     private StringSplitterInterface splitter = new StringSplitterInterface() {
         @Override
         public String[] split(String line) {
-            return line.split("[\\s]+");
+
+            // character tokenization for Chinese
+            if(ParametersForLbjCode.currentParameters.language.equals("zh")) {
```
mssammon: please define an enum or set of constant strings for the language codes

ner/src/main/java/edu/illinois/cs/cogcomp/ner/ExpressiveFeatures/GazetteersFactory.java
```diff
@@ -37,29 +43,35 @@
...
+        if(gazetteers_map.containsKey(path)) {
+            gazetteers = gazetteers_map.get(path);
+        } else {
+            synchronized (GAZ_INIT_LOCK) {
                 if (flatgazetteers) {
                     gazetteers = new FlatGazetteers(path);
                 } else {
                     gazetteers = new TreeGazetteers(maxPhraseLength, path);
                 }
-            } else {
```
mssammon: why change this? @cowchipkid any problems with this?

ner/src/main/java/edu/illinois/cs/cogcomp/ner/IO/ResourceUtilities.java
```diff
@@ -37,14 +37,7 @@ public static InputStream loadResource(String resourceFile) {
             // jar)
             if (!new File(file).exists()) {
                 logger.debug("Loading {} from classpath", resourceFile);
-                List<URL> list = IOUtils.lsResources(ResourceUtilities.class, resourceFile);
```
mssammon: if there is a problem with the IOUtils.lsResources() method, please create an issue and give an example

ner/src/main/java/edu/illinois/cs/cogcomp/ner/LbjTagger/Parameters.java
```diff
@@ -42,7 +42,7 @@
             "Affixes", "PreviousTag1", "PreviousTag2", "PreviousTagPatternLevel1",
             "PreviousTagPatternLevel2", "AggregateContext", "AggregateGazetteerMatches",
             "PrevTagsForContext", "PredictionsLevel1", "GazetteersFeatures", "WordEmbeddings",
-            "BrownClusterPaths", "Linkability"};
```
mssammon: If there is a dependency on Wikifier we may have a problem... does adding the flag result in any exceptions/problems if the source for these features is not present?

mssammon:
@danyaljj , @cowchipkid I'd appreciate your thoughts on this

ner/src/main/java/edu/illinois/cs/cogcomp/ner/ExpressiveFeatures/GazetteersFactory.java
cttsai: Because there might be language specific gazetteers, and sometimes different languages share the same set of gazetteers, we don't want to load them again.

ner/src/main/java/edu/illinois/cs/cogcomp/ner/ExpressiveFeatures/GazetteerTree.java
cttsai: ok. I have an enum of languages in [cross-lingual wikifier](https://github.com/cttsai/illinois-cross-lingual-wikifier/blob/master/src/main/java/edu/illinois/cs/cogcomp/xlwikifier/datastructures/Language.java). I can do the same here.

ner/src/main/java/edu/illinois/cs/cogcomp/ner/LbjTagger/Parameters.java
cttsai: These features are only populated by cross-lingual wikifier. There won't be any exception if they are not present.

cttsai: Regarding the size of language-specific resources: a language can have its own Brown clusters and gazetteers. Now I only have these for Chinese and Spanish. Other languages simply use the resources for English. The size of Brown clusters for one language is about 50M, and the gazetteers I added are about 100M.

ner/src/main/java/edu/illinois/cs/cogcomp/ner/ExpressiveFeatures/GazetteerTree.java
mssammon: I think it makes sense to add these somewhere like core-utilities so they are defined once

ner/src/main/java/edu/illinois/cs/cogcomp/ner/ExpressiveFeatures/GazetteersFactory.java
mssammon: OK -- this makes sense.  But it looks like you get rid of the option to use FlatGazetteers -- am I missing something?

ner/src/main/java/edu/illinois/cs/cogcomp/ner/ExpressiveFeatures/BrownClusters.java
```diff
@@ -62,42 +66,38 @@ private BrownClusters() {
...
+            if (ParametersForLbjCode.currentParameters.debug) {
+                logger.info(wordsAdded + " words added");
             }
+            brownclusters.wordToPathByResource.add(h);
+            brownclusters.isLowercaseBrownClustersByResource[i] =
+                    isLowercaseBrownClusters.elementAt(i);
+            brownclusters.resources.add(pathsToClusterFiles.elementAt(i));
+            in.close();
         }
     }
```
danyaljj: Doesn't it revert the changes made in #285 for synchronizing brown cluster access?

mssammon: Good catch. @cttsai please include the synchronization fix.

cttsai: Thanks. I didn't notice the difference before...

ner/src/main/java/edu/illinois/cs/cogcomp/ner/ExpressiveFeatures/GazetteerTree.java
cttsai: Is this a good place:
core-utilities/src/main/java/edu/illinois/cs/cogcomp/nlp/utilities

mssammon: better would be https://github.com/IllinoisCogComp/illinois-cogcomp-nlp/tree/master/core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/constants

ner/src/main/java/edu/illinois/cs/cogcomp/ner/IO/ResourceUtilities.java
cttsai: We had discussed this problem of reading a particular Chinese gazetteer in the classpath. This method cannot read all lines, but the method in LineIO can. I am trying to re-produce this issue, but the original method works fine somehow.... I'll remove my changes after some more tests.

ner/src/main/java/edu/illinois/cs/cogcomp/ner/config/NerBaseConfigurator.java
```diff
@@ -105,6 +106,7 @@
     private final static String DEFAULT_TREAT_ALL_FILES_AS_ONE = TRUE;
     private final static String DEFAULT_DEBUG = FALSE;
     private final static String DEFAULT_MODEL_NAME = "CoNLL";
+    private final static String DEFAULT_LANGUAGE = "en";
```
danyaljj: Do you want to add a comment here on the possible values for this field?

ner/src/main/lbj/LbjTagger.lbj
```diff
@@ -207,6 +219,19 @@ discrete% Affixes(NEWord word) <-
...
+	&& ParametersForLbjCode.currentParameters.language.equals("zh"))
+	{
+  		int N = word.form.length();
+	  	for (int i = 1; i <= 2; ++i)
+    		if (word.form.length() > i) sense "ZH-p|" : word.form.substring(0, i);
+  		for (int i = 1; i <= 2; ++i)
+    		if (word.form.length() > i) sense "ZH-s|" : word.form.substring(N - i);
+	}
+}
+
```
danyaljj: So don't we need to re-train/re-deploy the ner models because of adding this new features? 
(I'm actually surprised why none of the tests fail, since we are adding a new feature and the old model still works completely fine 😮  :thinking:  )

danyaljj: Oh this would be used only when the language is `ParametersForLbjCode.currentParameters.language.equals("zh")`; that's why it's not affected, I think ...

ner/src/main/java/edu/illinois/cs/cogcomp/ner/config/NerBaseConfigurator.java
cttsai: I'll add a enum of possible languages in core-utilities as Mark suggested.

Issue #147, opened 2016-08-10
inconsistent behavior in TextAnnotation.getTokenIdFromCharacterOffset()


mssammon:
If the last character offset in the text is used as an argument to this method, TextAnnotation returns one-past-the-end token index; this is  inconsistent with the method's behavior for any other index. Change to return the correct final token index.

cogcomp-dev:
please also write a relevant test.

mssammon:
@tkao1000 any questions about this?

tofuwen:
I did test like this:

@Test
    public void test() {
        // The construction of the John Smith library finished on time .
        TextAnnotation ta = DummyTextAnnotationGenerator.generateAnnotatedTextAnnotation(false, 1);
        System.out.println(ta);
        System.out.println(ta.getTokenIdFromCharacterOffset(54)); // the white space before "time"
        System.out.println(ta.getTokenIdFromCharacterOffset(60)); // .
        System.out.println(ta.getTokenIdFromCharacterOffset(61)); // whitespace after .
        // sentence length == 62
        // return the number of tokens + 1, as commented in the TextAnnotation.getTokenIdFromCharacter
        System.out.println(ta.getTokenIdFromCharacterOffset(62));
    }

And got the desired behavior. 

The "one-past-the-end token index" will occur when you pass the parameter as the length of the sentence, but I think that's what we want, since in the TextAnnotation.getTokenIdFromCharacterOffset, here is the comment for this:

        // If the characterOffset is the number of characters (and hence doesn't
        // correspond to any character with a zero indexed scheme), return the
        // number of tokens + 1.
        if (characterOffset == characterOffsetsToTokens.length) {
            return this.size();
        }
I think this is what you mean here.

mssammon:
This is good so far: please modify it so that instead of ```System.out``` you use ```assertEquals( <expectedValue>, ta.getTokenIdFrom...())```
Also, please check for other methods in TextAnnotation, View, and Constituent that use this method and make sure their behavior is as expected.

tofuwen:
No problem.
I commented here since I don't think I get inconsistent behaviors.
I will check other methods as well.

PR #295, opened 2016-11-28
Issue147

tofuwen:
#147 

Test for TextAnnotation.getTokenIdFromCharacterOffset

I am unable to find inconsistent behaviors for this method.

Also, in Constituent, View, TextAnnotation, no other method is calling this method.


mssammon:
The semaphore build is failing -- it looks like you need to add the license header to a new file.  You can do this with the command ```mvn license:format```

PR #304, opened 2016-12-13
Minor fixes to Pipeline README

bhargav:
- Update `PipelineFactory` class name.
- Add imports required to run the given code. [It is difficult to search for the actual namespaces while running]
- Remove dependencies that are not required for programmatic usage.

Pending:
The default set of configuration does not seem to work with pipeline. Will PR with correct default in a while.

### DIFF
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -96,10 +96,8 @@ of several other components for which it is a dependency.
 
 ### 1.3 LICENSE
 
-## Licensing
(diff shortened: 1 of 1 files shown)


pipeline/README.md
```diff
@@ -170,11 +170,6 @@ If this package is used in maven, please add the following dependencies with pro
 <dependencies>
```
danyaljj: could you add xml annotation to this snippet?

danyaljj: And java annotation to any java script which doesn't have any java annotations?

Issue #277, opened 2016-11-20
Moving illinois-quantifier to here? 


danyaljj:
Do we want to move the [illinois-quantifier](https://gitlab-beta.engr.illinois.edu/cogcomp/illinois-quantifier) to here?

danyaljj:
I have updated the quantifier package with the latest pipeline etc and deployed its models. Are people cool with moving it here? @mssammon are comments on this?

mssammon:
I'm OK with it.

PR #314, opened 2016-12-16
Quantifier

danyaljj:
Addresses #277 

Just moving it, no functional change. 

For future: add the quantifier annotator to the pipeline.


quantifier/src/main/java/edu/illinois/cs/cogcomp/quant/features/POSBigrams.java
```diff
@@ -0,0 +1,29 @@
...
+
+import edu.illinois.cs.cogcomp.core.datastructures.textannotation.Constituent;
+import edu.illinois.cs.cogcomp.edison.features.Feature;
+import edu.illinois.cs.cogcomp.edison.features.factory.SpanFeaturesOrdered;
+import edu.illinois.cs.cogcomp.edison.utilities.EdisonException;
+
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+public class POSBigrams extends LBJavaFeatureExtractor {
```
mssammon: A number of these feature extractors look like they may duplicated existing extractors in Edison.
The best way is probably to merge as is, then create an issue; probably requires some effort & testing to make sure there are no behavioral changes.

quantifier/src/main/java/edu/illinois/cs/cogcomp/quant/lbj/QuantitiesClassifier$$1.java
```diff
@@ -0,0 +1,91 @@
+/**
```
mssammon: do we want the lbjava-generated files to be part of the project?

quantifier/pom.xml
```diff
@@ -0,0 +1,87 @@
...
+            <version>4.11</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>edu.illinois.cs.cogcomp</groupId>
+            <artifactId>illinois-core-utilities</artifactId>
+            <version>3.0.87</version>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
```
mssammon: no problem with dependency on pipeline, if pipeline also ends up incorporating quantity extractor?

README.md
```diff
@@ -22,10 +22,10 @@ Each library contains detailed readme and instructions on how to use it. In addi
 | [illinois-lemmatizer](lemmatizer/README.md)  |  An application that uses [WordNet](https://wordnet.princeton.edu/) and simple rules to find the root forms of words in plain text. | 
 | [illinois-tokenizer](tokenizer/README.md) | An application that identifies sentence and word boundaries in plain text. | 
 | [illinois-pos](pos/README.md)  | An application that identifies the part of speech (e.g. verb + tense, noun + number) of each word in plain text.  |  
-| [illinois-ner](ner/README.md) | An application that identifies named entities in plain text according to two different sets of categories.  | 
+| [illinois-ner](ner/README.md) | An application that identifies named entities in plain text according to two different sets of categories.  |
+| [illinois-quantifier](quantifier/README.md) | This tool that etects mentions of quantities in the text, as well as normalizes it to a standard form. |
```
bhargav: nit: detects

danyaljj: Fixed.

quantifier/pom.xml
danyaljj: I think it's fine, as long as their versions are compatible (just like SRL).  In fact that's my next step: incorporating it in the pipeline.

quantifier/src/main/java/edu/illinois/cs/cogcomp/quant/lbj/QuantitiesClassifier$$1.java
danyaljj: I think these outputs are slightly modified after being generated from lbjava. But we can definitely change it so that it generates it on the fly, but we have to make sure about its functionality. How about checking this in like this, and delegating this to another issue?

danyaljj: BTW, except NER, all the lbjava-based modules in the project have their lbjava-based classes checked in (e.g. POS).

mssammon: We should think about this policy; I'll create an issue.  This module is fine for now, though.

quantifier/pom.xml
mssammon: Then we will know... :-)

PR #322, opened 2016-12-23
addimg more documentation for pipeline about webserver 

### LINKED ISSUE #313: Changes to the pipeline server
As suggested in #309 

- [ ] We should mention in the documentation that there is a light-weight REST API server available. 
- [x] Move the print statements to use a logger instance.
- [x] Add an API go get the names/information about views that the pipeline instance can populate. (This can be an extension in the future when we define the behavior properly)

danyaljj:
Closes #313

@bhargav @mssammon for either one of you.

### DIFF
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -221,7 +221,7 @@ Running your own text to get a visual sense of what IllinoisPreprocessor is doin
 scripts/runPreprocessor.sh  config/pipelineConfig.txt [yourTextFile] > [yourOutputFile]
 ```
 
-## 6. PROGRAMMATIC USE
+## 6. Programmatic Use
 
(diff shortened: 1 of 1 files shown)


pipeline/README.md
```diff
@@ -384,3 +384,24 @@ public class testpipeline {
     }
 }
 ```
+
+## 7. Using pipeline webserver 
+
+Our pipeline contains a webserver wich can be run on a remote server. The server supports post and get requests to obtain annotation for a requested text, with desired views.
```
bhargav: Typo: `which`.

Also we should mention that it is a very *basic* webserver.

danyaljj: Updated.

PR #333, opened 2017-01-05
adding subTextAnnotation extraction.

danyaljj:
Adding a method to `TextAnnotationUtilities` to extract a subset of a TextAnnotation object.


core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/datastructures/textannotation/TextAnnotationUtilities.java
```diff
@@ -23,79 +25,61 @@
...
+            (arg0, arg1) -> {
+                int size0 = arg0.size();
+                int size1 = arg1.size();
+
+                if (size0 < size1)
+                    return -1;
+                else if (size0 > size1)
+                    return 1;
+                else
+                    return 0;
```
danyaljj: Everything above here, is just simplifications using lambdas.

PR #335, opened 2017-01-07
adding abbreviation extractor

danyaljj:


edison/src/test/java/edu/illinois/cs/cogcomp/edison/features/TestAbbreviationExtractor.java
```diff
@@ -0,0 +1,44 @@
...
+    private static AbbreviationFeatureExtractor abbreviationFeatureExtractor = new AbbreviationFeatureExtractor();
+
+    @Test
+    public void testAbbreviations() throws EdisonException {
+        try {
+            abbreviationFeatureExtractor.addView(ta);
+        } catch (AnnotatorException e) {
+            e.printStackTrace();
+        }
+        assert ta.getAvailableViews().contains(abbreviationFeatureExtractor.getViewName());
```
mssammon: ideally, add at least one test for an expected abbreviation in the text, in a way that shows how a user would retrieve the relevant information from the TextAnnotation.

edison/src/test/java/edu/illinois/cs/cogcomp/edison/features/TestAbbreviationExtractor.java
```diff
@@ -0,0 +1,44 @@
...
+        assert ta.getView(abbreviationFeatureExtractor.getViewName()).getConstituents().size() >= 1;
+    }
+}
```
mssammon: Given the heavy use of parentheses in the abbreviation identifier,  add a test case with multiple parens -- including unbalanced parentheses -- to demonstrate that it is robust.

PR #337, opened 2017-01-08
fixing the lazy initialization for the quantifier.

### LINKED ISSUE #327: Lazy initialization for quantifier
Verify that the initialization of [the quantifier annotator](https://github.com/IllinoisCogComp/illinois-cogcomp-nlp/blob/master/quantifier/src/main/java/edu/illinois/cs/cogcomp/quant/driver/Quantifier.java) is properly done in a "lazy" way.

danyaljj:
Closes #327

### DIFF
--- a/quantifier/src/main/java/edu/illinois/cs/cogcomp/quant/driver/Quantifier.java
+++ b/quantifier/src/main/java/edu/illinois/cs/cogcomp/quant/driver/Quantifier.java
@@ -149,9 +149,9 @@ public class Quantifier extends Annotator {
             if (!prediction.equals("O")
                     && tokenPos < annotation.size()
                     && (i == (tokens.size() - 1)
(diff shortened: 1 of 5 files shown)


quantifier/src/main/java/edu/illinois/cs/cogcomp/quant/driver/Quantifier.java
```diff
@@ -44,7 +44,9 @@
     }
 
     public Quantifier() {
-        super(ViewNames.QUANTITIES, new String[0]);
+        // lazily initialize by default.
+        this(true);
+
         normalizer = new Normalizer();
```
mssammon: shouldn't these go either in the initalize() method or the other constructor?

danyaljj: done.

PR #341, opened 2017-01-17
ERE reader

mssammon:
Improved ERE NER reader. 
Added modes to build just NER view, add nominal and filler mentions.
Added Relation and Coreference readers. 
Added a non-unit test to check behaviors.


danyaljj:
Had a light skim through the changes and it looks good to me. Hopefully @cowchipkid and  @bhargav will have a deeper look into it.

corpusreaders/src/main/java/edu/illinois/cs/cogcomp/nlp/corpusreaders/ereReader/EREDocumentReader.java
```diff
@@ -7,42 +7,162 @@
...
+ * @author redman
+ * @author msammon
  */
 public class EREDocumentReader extends XmlFragmentWhitespacingDocumentReader {
-	
-	/** these tags contain attributes we want to keep. */
-    static private ArrayList<String> retainTags = new ArrayList<String>();
+
+    /**
+     * tags in ERE markup filess
```
bhargav: Minor: These tags can be private to the class. We might not need them outside.

corpusreaders/src/main/java/edu/illinois/cs/cogcomp/nlp/corpusreaders/ereReader/EREDocumentReader.java
```diff
@@ -7,42 +7,162 @@
...
+    public static final String ARG_ONE = "rel_arg1";
+    public static final String ARG_TWO = "rel_arg2";
+    public static final String ENTITY_MENTION_ID = "entity_mention_id";
+    public static final String ENTITY_ID = "entity_id";
+    public static final String ROLE = "role";
+    public static final String FILLER_ID = "filler_id";
+
+
+
+    /** aim for consistent naming */
```
bhargav: For more consistency, we should try to move the constants to some place independent of the reader. This can be part of some future refactoring.

corpusreaders/src/main/java/edu/illinois/cs/cogcomp/nlp/corpusreaders/ereReader/EREMentionRelationReader.java
```diff
@@ -0,0 +1,173 @@
...
+ * Developed by: The Cognitive Computation Group University of Illinois at Urbana-Champaign
+ * http://cogcomp.cs.illinois.edu/
+ */
+package edu.illinois.cs.cogcomp.nlp.corpusreaders.ereReader;
+
+import edu.illinois.cs.cogcomp.core.datastructures.Pair;
+import edu.illinois.cs.cogcomp.core.datastructures.ViewNames;
+import edu.illinois.cs.cogcomp.core.datastructures.textannotation.*;
+import edu.illinois.cs.cogcomp.nlp.corpusreaders.aceReader.SimpleXMLParser;
+import edu.illinois.cs.cogcomp.nlp.corpusreaders.aceReader.XMLException;
```
bhargav: We should try to move these XML parsers outside the ACE Reader namespace. Future refactoring maybe.

corpusreaders/src/main/java/edu/illinois/cs/cogcomp/nlp/corpusreaders/AbstractIncrementalCorpusReader.java
```diff
@@ -30,8 +31,12 @@
...
-    // list of all files that comprise corpus
-    private List<Path> fileList;
+    /**
+     * contains pointers to files comprising corpus.  Each entry may consist of a source document (first element
+     *    of each list) plus zero or more files with related annotations needed to instantiate the
+     *    corresponding TextAnnotation objects.
+     */
+    private List<List<Path>> fileList;
 
     // holds TextAnnotations extracted from most recently accessed file
```
cowchipkid: I notice the field above has a corrected JavaDoc comment included(thanks Mark), I suggest if we do it for one, we do it for all. In fact, we should force a JavaDoc style sheet. That's another conversation.

corpusreaders/src/main/java/edu/illinois/cs/cogcomp/nlp/corpusreaders/ereReader/EREDocumentReader.java
mssammon: This is a vexed issue, because different corpora almost invariably have overlapping but non-identical tag sets.  I opted here to have a local copy of all the variables, and to reuse those where they were the same in an attempt to help anyone who is moving between ERE and ACE to identify where the tags and attributes are the same and where they are different -- it affects what feature extractors can be reused, for example.

mssammon:
If no one else merges, I will do so later today... @danyaljj @bhargav @cowchipkid

PR #344, opened 2017-01-18
a method to return constitutnes in a given char span.

danyaljj:
Added a method to View to return constituents, given char-begin-offset and char-end-offset.


danyaljj:
> Is this operation likely to be called very frequently? If so, we should consider storing a temporary map from char offsets to Constituents
I'm not sure how frequently this will be used, but adding a map would be helpful, if it turns out useful (future work). 

 > In javadoc, indicate this uses one-past-the-end indexing as elsewhere in Constituent/TextAnnotation.

Added the javadoc comment btw.

danyaljj:
Addressed the comments. Planning to merge when it turns green.

PR #348, opened 2017-01-20
upgrade stanford corenlp version 

### LINKED ISSUE #289: Update ver. of Stanford deps in the pipeline?
The current versions are 3.3.1 which were release about 3 yrs ago. Maybe we upgrade to their latest version? (clearly, before doing this, we have to have at least a couple of tests passing to verify that this transition is done correctly).

danyaljj:
Closes #289 

FYI @christos-c @microth

### DIFF
--- a/pipeline/src/test/java/edu/illinois/cs/cogcomp/pipeline/main/CachingPipelineTest.java
+++ b/pipeline/src/test/java/edu/illinois/cs/cogcomp/pipeline/main/CachingPipelineTest.java
@@ -162,12 +162,17 @@ public class CachingPipelineTest
             System.out.println("Expected exception from stanford.");
         }
         String predictedDepTree = basicTextAnnotation.getView(ViewNames.DEPENDENCY_STANFORD).toString();
-        String goldDepTree = "(taken (:LABEL:prep In (:LABEL:pobj States :LABEL:det the\n" +
-                "                    :LABEL:nn United))\n" +
-                "       (:LABEL:nsubj Cinco (:LABEL:prep de :LABEL:pobj Mayo))\n" +
+        String goldDepTree = "(taken :LABEL:punct .\n" +
+                "       (:LABEL:nmod States :LABEL:case In\n" +
+                "                           :LABEL:det the\n" +
+                "                           :LABEL:compound United)\n" +
+                "       :LABEL:punct ,\n" +
+                "       (:LABEL:nsubj Cinco (:LABEL:nmod Mayo :LABEL:case de))\n" +
                 "       :LABEL:aux has\n" +
-                "       (:LABEL:prep on (:LABEL:pobj significance :LABEL:det a))\n" +
-                "       (:LABEL:prep beyond (:LABEL:pobj that (:LABEL:prep in :LABEL:pobj Mexico))))";
+                "       (:LABEL:nmod significance :LABEL:case on\n" +
+                "                                 :LABEL:det a)\n" +
+                "       (:LABEL:nmod that (:LABEL:nmod Mexico :LABEL:case in)\n" +
+                "                         :LABEL:case beyond))";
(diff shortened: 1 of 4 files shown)


pipeline/src/test/java/edu/illinois/cs/cogcomp/pipeline/main/CachingPipelineTest.java
```diff
@@ -162,12 +162,17 @@ public void stanfordParseHandler() {
...
+        String goldDepTree = "(taken :LABEL:punct .\n" +
+                "       (:LABEL:nmod States :LABEL:case In\n" +
+                "                           :LABEL:det the\n" +
+                "                           :LABEL:compound United)\n" +
+                "       :LABEL:punct ,\n" +
+                "       (:LABEL:nsubj Cinco (:LABEL:nmod Mayo :LABEL:case de))\n" +
                 "       :LABEL:aux has\n" +
-                "       (:LABEL:prep on (:LABEL:pobj significance :LABEL:det a))\n" +
-                "       (:LABEL:prep beyond (:LABEL:pobj that (:LABEL:prep in :LABEL:pobj Mexico))))";
+                "       (:LABEL:nmod significance :LABEL:case on\n" +
```
mssammon: Here, and for 'de' above, the dependency tree has the label "case" where it used to be "prep". This is apparently due to the shift to universal dependencies.  Changes like this may break (or at least render inoperative) a good number of feature extractors, so we should make it very clear to anyone upgrading from the old i-c-n to the new one.  Also, we will have to retrain SRL for the pipeline to work with this new version.

quantifier/src/main/java/edu/illinois/cs/cogcomp/quant/driver/QuantDemo.java
```diff
@@ -41,6 +41,7 @@ public static String getJSONFromRatio(Ratio r) {
 
     public static void main(String args[]) throws Exception {
         Quantifier quantifier = new Quantifier();
+        quantifier.doInitialize();
```
mssammon: Is this necessary?

danyaljj: Orthogonal to this PR. This was missing from the previous PRs.

christos-c:
@mssammon Concerning SRL's performance, the new DNN parser from Stanford (if this is the default one in this release) _should_ be better than the old one, but of course, we should try re-parsing all of Propbank and retain/retest to see the impact. Personally, I'd be much happier to move SRL to a dependency-only solution with @mliu-dark-knight's parser.

danyaljj:
I think it will def affect the performance of the SRL. 
I should make the Stanford dependency, optional so that it doesn't affect the SRL. 
Does that sound any good, @mssammon ?

mssammon:
probably we should remove the dependency in pipeline, so that it comes via the SRL.  Sound right?

danyaljj:
Nope. I want to keep the new version and keep SRL using the old one.

mssammon:
Right now, I believe the SRL takes as input the data structure populated by the previous steps in the pipeline.  So how will your solution allow the behavior you want?

danyaljj:
@mssammon we can use srl's internal pre-processing by turning [this](https://github.com/IllinoisCogComp/illinois-cogcomp-nlp/blob/cc3a5d027b659ea34ef4818cbc8da20a0af4d31f/pipeline/src/main/java/edu/illinois/cs/cogcomp/pipeline/common/PipelineConfigurator.java#L47-L48) on; that should be fine. Right?

mssammon:
Yes, if the dependency conflict is resolved (e.g. with maven shade plugin)

danyaljj:
Ah yea you're right.

PR #356, opened 2017-01-22
bug fix in ERE reader

mssammon:
Mentions of type 'filler' were being created with empty surface forms, resulting in errors downstream. Added a fix and an explicit check in the reader.


pipeline/src/main/java/edu/illinois/cs/cogcomp/pipeline/server/MainServer.java
```diff
@@ -29,7 +29,7 @@
 
 public class MainServer {
 
-    private static Logger logger = LoggerFactory.getLogger(StanfordDepHandler.class);
```
danyaljj: 👍

danyaljj:
Considering that the changes are mostly small, I'm merging this. @bhargav let us know if you noticed any issues.

PR #365, opened 2017-02-08
change the module names 

danyaljj:
Pattern: 
 - illinois-x ---> cogcomp-x 
 - edu.illinois.cs.cogcomp --> edu.cs.cogcomp


christos-c:
I second that, it's going to cause major backwards-compatibility problems.

PR #369, opened 2017-02-18
updates to corpusreaders

mssammon:
Updated corpus reader interface to include a method to generate a human-readable report.  For most readers, this will simply throw an exception indicating it is not yet implemented.
Fixes to ERE-related readers. 
Some minor formatting changes.


corpusreaders/src/main/java/edu/illinois/cs/cogcomp/nlp/corpusreaders/ereReader/EREMentionRelationReader.java
```diff
@@ -174,4 +194,26 @@ public String getViewName() {
...
+     * Reports number of relations and relation mentions read from source and generated. Note that
+     * number read and number generated may disagree because an argument was not read correctly (see
+     * {#ERENerReader.generateReport()})
+     * @return a String representing a human-readable report on information read from corpus and
+     *         generated as data structures.
+     */
+    @Override
+    public String generateReport() {
+        StringBuilder bldr = new StringBuilder(super.generateReport());
+        bldr.append("Number of relations in source: ").append(numRelationsInSource)
```
bhargav: Minor: Do we need to set these counters to 0 when the `.reset()` method on the Document Reader is called and the dataset is read again (I am not sure if all reader cache the dataset in memory and not parse multiple times -- in which case, reset does not cause problems with counters)?

danyaljj:
LGTM.

corpusreaders/src/main/java/edu/illinois/cs/cogcomp/nlp/corpusreaders/ereReader/EREMentionRelationReader.java
mssammon: good point -- I updated classes in the inheritance stack.

PR #376, opened 2017-02-28
adding the pipeline client annotator.

### LINKED ISSUE #375: [pipeline-server] add a client for pipeline server
Add an option for the pipeline annotator to choose send requests to the pipeline server.

danyaljj:
Closes #375 

I also did a code style formatting which cluttered the overall diff.  But the main change is easier to see in the first commit.

### DIFF
--- a/pipeline/src/main/java/edu/illinois/cs/cogcomp/pipeline/common/HandlerUtils.java
+++ b/pipeline/src/main/java/edu/illinois/cs/cogcomp/pipeline/common/HandlerUtils.java
@@ -16,26 +16,22 @@ import java.util.List;
 /**
  * Created by mssammon on 2/1/16.
  */
-public class HandlerUtils
-{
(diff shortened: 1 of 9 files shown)


pipeline/src/main/java/edu/illinois/cs/cogcomp/pipeline/common/HandlerUtils.java
```diff
@@ -47,16 +43,17 @@ public static Constituent checkTextAnnotationRespectsSentenceLengthLimit(TextAnn
 
     /**
      * generate an error message for sentence length error.
+     * 
      * @param taId Id of text from which sentence came
-     * @param sentence  sentence text
-     * @param maxSentenceLength  length limit
```
mssammon: is there a good way to factor out formatting changes from the change set?

PR #380, opened 2017-03-01
adding caching to server client.

danyaljj:
- added the possibility of caching on the server-client's side.


danyaljj:
Applied the change: "System.out -> logger."

Merging.

PR #381, opened 2017-03-02
Update Gurobi dependency to 7.0.1.

bhargav:
- Additionally Jenkins on Morgoth has been updated to use Gurobi 7.0.1


danyaljj:
Something in the [dependency parse is failing](http://morgoth.cs.illinois.edu:8080/job/cogcomp-nlp-pr/301/edu.illinois.cs.cogcomp$illinois-depparse/console), but most-likely it is not related to this PR, since dependency parse is not using Gurobi. But we have to keep eyes on it.

PR #383, opened 2017-03-02
[core-utils] addView exception handled

danyaljj:
Sometimes a single annotator fails, in which case we shouldn't lose the whole text-annotation and the rest of the views. So here I handled the exception thrown by a single `addView` functions of the annotators.

### DIFF
--- a/core-utilities/src/main/java/edu/illinois/cs/cogcomp/annotation/BasicAnnotatorService.java
+++ b/core-utilities/src/main/java/edu/illinois/cs/cogcomp/annotation/BasicAnnotatorService.java
@@ -371,7 +371,12 @@ public class BasicAnnotatorService implements AnnotatorService {
                 ta = annotationCache.getTextAnnotation(ta);
 
         for (String viewName : viewsToAnnotate) {
-            isUpdated = addView(ta, viewName) || isUpdated;
+            try {
+                isUpdated = addView(ta, viewName) || isUpdated;
+            } catch (AnnotatorException e) {
+                // the exception is handled here, because one single view failure should not resutl in loss of all the annotations
+                e.printStackTrace();
+            }
         }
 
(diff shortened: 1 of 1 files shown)


core-utilities/src/main/java/edu/illinois/cs/cogcomp/annotation/BasicAnnotatorService.java
```diff
@@ -371,7 +371,12 @@ public TextAnnotation addViewsAndCache(TextAnnotation ta, Set<String> viewsToAnn
                 ta = annotationCache.getTextAnnotation(ta);
 
         for (String viewName : viewsToAnnotate) {
-            isUpdated = addView(ta, viewName) || isUpdated;
+            try {
+                isUpdated = addView(ta, viewName) || isUpdated;
+            } catch (AnnotatorException e) {
+                // the exception is handled here, because one single view failure should not resutl in loss of all the annotations
```
bhargav: Personally, I do not like this being the default behavior for handling errors. We should defined a runtime parameter for the user to explicitly choose this behavior. Let me know what you think?

Aside: This will also affect downstream MapDB caching if the TA is cached using hash of text, corpus Id, or incorrect set of views.

mssammon: I agree that this should be configurable. But I think default should be "don't fail". 
Regarding downstream caching: I think the current approach works (text hash is used just to retrieve existing annotations; individual views are checked each time), though perhaps not for sentence-by-sentence.

bhargav: Okay. Makes sense. We should log the exception as an error.

danyaljj: done.

PR #389, opened 2017-03-06
brownCluster using datastore

danyaljj:
Replaced maven-dependency with datastore. 
If you think this is good, I can apply the same thing to the rest of the resources. 

-------------  
Side note: if you wanna see where these files are stored, login to minio, using the credentials I sent you last week: http://smaug.cs.illinois.edu:8080/minio/

You should be able to see this: 
screen shot 2017-03-05 at 10 44 34 pm
 On the left we see two main buckets: one for private files (not accessible for public, like copy-righted files, training data, etc) and read-only files (only readable for public, like Edison resources). And you can see that inside the read-only files we have the brown-clusters, which are being used here. 

You can upload either from this graphical interface, or use the `datastore` programmatically. 

FYI @shyamupa


edison/src/main/java/edu/illinois/cs/cogcomp/edison/annotators/BrownClusterViewGenerator.java
```diff
@@ -107,6 +106,24 @@ private void loadClusters() throws Exception {
         if (gzip)
             stream = new GZIPInputStream(stream);
 
+        loadFromInputStream(stream);
+    }
+
+    public void loadFromDataStore() throws Exception {
+        Datastore dsNoCredentials = null;
+        try {
+            dsNoCredentials = new Datastore("http://smaug.cs.illinois.edu:8080");
```
danyaljj: Possibly for future, I think we should add a `ResourcesConfigurator` to the core-utils to contain these details.

mssammon: agreed. Can we add one in this PR?

bhargav:
One question regarding versioning. Current folder structure is flat with version appended at the end of the file. How about `version/<versioned_files>` structure like maven?

edison/src/main/java/edu/illinois/cs/cogcomp/edison/annotators/BrownClusterViewGenerator.java
```diff
@@ -12,18 +12,16 @@
 import edu.illinois.cs.cogcomp.core.datastructures.ViewNames;
 import edu.illinois.cs.cogcomp.core.datastructures.textannotation.SpanLabelView;
 import edu.illinois.cs.cogcomp.core.datastructures.textannotation.TextAnnotation;
-import edu.illinois.cs.cogcomp.core.datastructures.textannotation.View;
 import edu.illinois.cs.cogcomp.core.io.IOUtils;
 import edu.illinois.cs.cogcomp.core.utilities.StringUtils;
 import edu.illinois.cs.cogcomp.core.utilities.configuration.ResourceManager;
 import edu.illinois.cs.cogcomp.edison.config.BrownClusterViewGeneratorConfigurator;
+import io.minio.errors.*;
```
bhargav: IMO we should have datastore throw its own Exception instances so that the client does not have to deal with Minio directly. Right now, this assumes that the minio sdk (and namespaces) would always be available transitively. Maybe for a future PR.

edison/src/main/java/edu/illinois/cs/cogcomp/edison/annotators/BrownClusterViewGenerator.java
```diff
@@ -107,6 +106,24 @@ private void loadClusters() throws Exception {
...
+        Datastore dsNoCredentials = null;
+        try {
+            dsNoCredentials = new Datastore("http://smaug.cs.illinois.edu:8080");
+            System.out.println(file);
+            File f = dsNoCredentials.getFile("edu.cogcomp", file, 1.3);
+            InputStream is = new FileInputStream(f);
+            loadFromInputStream(is);
+        } catch (InvalidPortException | InvalidEndpointException e) {
+            e.printStackTrace();
+            System.out.println("The BrownCluster did not load. This is probably because out datastore server is off. Try again later or ping us. ");
```
bhargav: logger.error? (Also logger.error lets you provide the exception instance as an argument. So that appropriate information is shown in the logs.) https://www.slf4j.org/api/org/slf4j/Logger.html#error(java.lang.String, java.lang.Throwable)

edison/src/main/java/edu/illinois/cs/cogcomp/edison/annotators/BrownClusterViewGenerator.java
danyaljj: done.

edison/src/main/java/edu/illinois/cs/cogcomp/edison/annotators/BrownClusterViewGenerator.java
danyaljj: Yup. Will create a DatastoreAccessException or sth.

danyaljj:
Updated the PR with moving all the resources to the datastore. 
Some resources (like Gazetteers or Roget, didn't have a unit test and they were not functional). I added the tests and made sure they work. 
Some resources (#393) are never used, so I dropped them.

mssammon:
Added resources *and* unit tests. You are my hero.

mssammon:
@danyaljj @bhargav any reason not to go ahead with this?

danyaljj:
It's all me slacking off. I will finish this and let you know; definitely before end of this week.

mssammon:
no rush -- I just wanted to make sure I'm not the bottleneck here...

danyaljj:
Somehow Semaphore is not triggered on this. Closing it to open a new one.

Issue #391, opened 2017-03-08
What to do with the MapDB cache when program fails?


danyaljj:
We use MapDB extensively in our work. The issue that comes up often is that, when the program fails/crashes, the cache doesn't seem to be valid anymore. 
Like on the next run it says:
```
org.mapdb.DBException$DataCorruption: Header checksum broken. Store was not closed correctly, or is corrupted
```

Is there a way to handle cache safely?

PR #392, opened 2017-03-08
adding transaction to server client

danyaljj:
Partially addressing #391: 
https://jankotek.gitbooks.io/mapdb/content/performance/

### DIFF
--- a/pipeline/src/main/java/edu/illinois/cs/cogcomp/pipeline/server/ServerClientAnnotator.java
+++ b/pipeline/src/main/java/edu/illinois/cs/cogcomp/pipeline/server/ServerClientAnnotator.java
@@ -82,7 +82,10 @@ public class ServerClientAnnotator extends Annotator {
     }
 
     public void useCaching(String dbFile) {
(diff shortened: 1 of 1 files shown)


pipeline/src/main/java/edu/illinois/cs/cogcomp/pipeline/server/ServerClientAnnotator.java
```diff
@@ -82,7 +82,7 @@ public void useCaching() {
     }
 
     public void useCaching(String dbFile) {
-        this.db = DBMaker.fileDB(dbFile).closeOnJvmShutdown().make();
+        this.db = DBMaker.fileDB(dbFile).closeOnJvmShutdown().transactionEnable().make();
```
bhargav: I think you will also have to call `this.db.commit()` whenever you make an update to the data. 

https://jankotek.gitbooks.io/mapdb/content/db#transactions

danyaljj: Good catch.

bhargav:
Looks good.

PR #396, opened 2017-03-11
incorporating NESim into package

lingss0918:
move NESim from gitlab: https://gitlab.engr.illinois.edu/cogcomp/illinois-entity-similarity into the similarity package.


bhargav:
Some quick comments. Didn't review the entire PR.

1. Some files are not formatted properly. There should be some tool to format Java code (@danyaljj might know).
2. You are missing dependencies in `similarity/pom.xml`. Especially for classes - `com.wcohen.secondstring.JaroWinkler` and `org.apache.xmlrpc.client.XmlRpcClient`.
3. The `StandardInputOutput` class contains methods that have equivalent ones in IOUtils and LineIO. We should try to avoid duplicated behavior here and maybe move new functions to core-utilities.

Will review thoroughly in sometime later.

danyaljj:
For 1: you can run `mvn -pl similarity formatter:format`.

PR #397, opened 2017-03-15
parse xml, clean text, retain original offsets

mssammon:
This code has two components: 
1. StringTransformation: allows users to clean text but maintain a mapping to the original text offsets. This allows substitution sequences that are longer or shorter than the originals as well as deletions.
2. XmlTextAnnotation: parse xml corpus files into TextAnnotations with clean text amenable to NLP processing (using StringTransformation to map to original xml file offsets), plus xml annotations (like date/time, post author etc.).  This latter allows e.g. coreference systems to use post authors as mentions, per ERE corpus specs.


core-utilities/src/main/java/edu/illinois/cs/cogcomp/annotation/XmlTextAnnotationMaker.java
```diff
@@ -0,0 +1,74 @@
...
+ *    offsets) is also derived.
+ * The goal is to provide text that can be processed easily with an NLP pipeline without a lot of hacks to work around
+ *    ill-formatted text. The annotations so produced can then be mapped to the offsets in the original xml text,
+ *    and combined with supplementary information extracted from the xml markup.
+ *
+ * @author mssammon
+ */
+
+public class XmlTextAnnotationMaker {
+
```
cowchipkid: What are the rules for the use of blank lines? Where do we use one, where do we use two and where do we use three? And why?

core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/datastructures/textannotation/TextAnnotationUtilities.java
```diff
@@ -82,7 +82,14 @@ else if (size0 > size1)
                 else
                     return 0;
             };
+    public static final Comparator<IntPair> IntPairComparator = (arg0, arg1) -> {
```
cowchipkid: I suspect I missed something, do we now require Java 8?

core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/utilities/StringTransformation.java
```diff
@@ -0,0 +1,312 @@
...
+ * These transformations modify the character offsets of the remaining string (e.g. replacing xml-escaped
+ *     characters with their ascii counterparts, without padding the replacement with whitespace).
+ * This could include deletion of markup, in which case information associated with the markup might be stored.
+ *
+ * @author mssammon
+ */
+public class StringTransformation {
+
+    private final String origText;
+    // tracks whether transformedText field is out of date
```
cowchipkid: Why are these not in proper JavaDoc format? Rather than the "//", use "/**" to open and "*/" to close, the comment will be added to the javadocs...

core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/utilities/TextCleanerStringTransformation.java
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    private boolean removeRepeatPunctuation;
+    private boolean replaceUnderscores;
+    private boolean replaceControlSequence;
+    private boolean replaceAdHocMarkup;
+    private boolean replaceBogusApostrophe;
+
+
+    public TextCleanerStringTransformation(ResourceManager rm_) throws SAXException, IOException {
```
cowchipkid: Why is there an underscore after the argument name (rm_)?

core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/utilities/TextCleanerStringTransformation.java
```diff
@@ -0,0 +1,254 @@
...
+            int start = matcher.start();
+            int end = matcher.end();
+
+            origTextSt.transformString(start, end, replacement);
+        }
+
+        return origTextSt;
+    }
+
+    public static String replaceXmlTags(String origText) {
```
cowchipkid: I think we need some javadoc on this method.

core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/utilities/XmlDocumentProcessor.java
```diff
@@ -0,0 +1,159 @@
...
+ *    and others as metadata.
+ * Provides a mapping from clean text character offsets to character offsets in the source xml.
+ * Metadata includes relevant character offsets.
+ *
+ * TODO: add constructor field to allow additional text clean/transform ops
+ */
+
+public class XmlDocumentProcessor {
+
+    private final Set<String> tagsWithText;
```
cowchipkid: These fields need javadocs.

core-utilities/src/main/java/edu/illinois/cs/cogcomp/annotation/XmlTextAnnotationMaker.java
```diff
@@ -0,0 +1,74 @@
...
+    private final TextAnnotationBuilder taBuilder;
+    private final XmlDocumentProcessor xmlProcessor;
+
```
cowchipkid: What are the rules for the use of blank lines? Where do we use one, where do we use two and where do we use three? And why?

core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/datastructures/textannotation/TextAnnotationUtilities.java
mssammon: Yes -- see the main project pom.xml file (see `<source>` element).

core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/utilities/TextCleanerStringTransformation.java
mssammon: historical artifact.

core-utilities/src/main/java/edu/illinois/cs/cogcomp/annotation/XmlTextAnnotationMaker.java
mssammon: My intent is to follow the guidelines on vertical spacing here:
https://google.github.io/styleguide/javaguide.html#s4.6.1-vertical-whitespace
this has a lot of flexibility in terms of using empty lines to group related content.  My goal is to make the code as readable -- to myself, anyway -- as possible.   But sometimes my `return` finger is a bit itchy.

cowchipkid: I agree sparing with the use of single returns used strategically as outlined in the style guide. But throughout the code base, I see one, two and three spaces used. I assume every character, punctuation and newline has meaning in a source file(which is why I questioned the argument named "rm_"). I suggest we better define the conventions concerning use of blank lines, I think a very simple policy would eliminate guesswork on this.

mssammon:
Agree with your concern, but full xml parser is out of scope...

PR #401, opened 2017-03-23
NER Parameter Tuning

cowchipkid:
This collection of scripts in total does a parameter sweep over learning rate and thickness.


danyaljj:
Do we want to checkin `ner/scripts/paramsweep/run.jar` into the repo?

cowchipkid:
Yes. I no longer have the source that little app I wrote.

mssammon:
I'd prefer non-binary -- even if it were just decompiled and cleaned up a bit.... how much work would that be?

cowchipkid:
It would be faster to rewrite it. Might be able to do this in bash, but I am not that well versed bash… If we want to put that much labor into, I would also suggest we change it such that it no longer programmatically modifies and recompiles the LBJ file, but rather simply modifies the config files, adding the appropriate parameters there.

mssammon:
any idea how much time it would take (in Java)?

cowchipkid:
1 - 2 hrs to write, but then would need to be tested, but I would be hands off at that point.

mssammon:
definitely worth rewriting, then.  Please go ahead and add the code to a suitable package in the cogcomp-nlp project.

cowchipkid:
Moving the illinois-cogcomp-nlp to cogcomp-nlp has broken the scripts anyway. I will fix those as well.

ner/src/main/java/edu/illinois/cs/cogcomp/ner/ParameterSweep.java
```diff
@@ -0,0 +1,111 @@
+/**
```
mssammon: need to add license header -- `mvn license:format`

PR #407, opened 2017-03-31
adding view names as arguments of buildPipeline.

### LINKED ISSUE #288: Simplify the pipeline usage?
```
// creates a StanfordCoreNLP object, with POS tagging, lemmatization, NER, parsing, and coreference resolution 
Properties props = new Properties();
props.setProperty("annotators", "tokenize, ssplit, pos, lemma, ner, parse, dcoref");
StanfordCoreNLP pipeline = new StanfordCoreNLP(props);

// read some text in the text variable
String text = ... // Add your text here!

// create an empty Annotation just with the given text
Annotation document = new Annotation(text);

// run all Annotators on this text
pipeline.annotate(document);
```

And here is what we give to the user: 

```
String docId = "APW-20140101.3018"; // arbitrary string identifier
String textId = "body"; // arbitrary string identifier
String text = ...; // contains plain text to be annotated

ResourceManager rm = new ResourceManager( "config/pipeline-config.properties" );
AnnotatorService pipeline = IllinoisPipelineFactory.buildPipeline( rm );
TextAnnotation ta = pipeline.createAnnotatedTextAnnotation( docId, textId, text );
```

The thing that annoys me is that this whole `ResourceManager` business can be confusing to new users. Is it possible to have an additional/alternative method without the need to create a `ResourceManager`? For example you just say 

```
AnnotatorService pipeline = IllinoisPipelineFactory.buildPipeline("chunk", "pos");
```

or 

```
AnnotatorService pipeline = IllinoisPipelineFactory.buildPipeline("srl");
```

danyaljj:
Closes #288

### DIFF
--- a/pipeline/README.md
+++ b/pipeline/README.md
@@ -3,24 +3,6 @@
 This module contains code that allows you to run various NLP tools
 to annotate plain text input. 
 
-## CONTENTS
-0. Quickstart
-1. Purpose
-	1. Intended Use
-	2. Pipeline NLP Components
-	3. License
-2. Prerequisites
-	1.  General Requirements: System
-	2.  Specific Requirements: SRL
-3. Download contents 
-4. Dependencies
-5. Running the Illinois NLP Pipeline
-	1.  Running a Simple Command-Line Test
-6. Programmatic use
-	1. Configuration Options
-	2. Changing the logging settings
-	3. Troubleshooting
-
 ## 0. Quickstart
 
 Assuming you downloaded the pipeline package as a zip from the
@@ -34,7 +16,7 @@ scripts/runPipelineOnDataset.sh  <configFile> <inputDirectory> <outputDirectory>
 The configuration file needs only to contain options to override defaults.
 
 
-## 1. Purpose
+## Purpose
 
 This software bundles some basic preprocessing steps that a lot of NLP
 applications need, with the goal of making them run locally. Some
@@ -52,7 +34,7 @@ the same data set multiple times), it will use a cached copy of the
 system outputs and run much faster.
 
 
-### 1.1 Intended use
+### Intended use
 
 The illinois-nlp-pipeline package was designed to be used either
 programmatically -- inline in your Java code -- or from the command line,
@@ -70,7 +52,7 @@ illinois-core-utilities, and use it as an argument to a
 TokenizerTextAnnotationBuilder (also from illinois-core-utilities).
 
 
-## 1.2 Pipeline NLP Components
+### Pipeline NLP Components
 
 The pipeline has the following annotators. To understand the annotations,
 please refer to the descriptions of the individual packages at the URLs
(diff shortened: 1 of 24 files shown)


pipeline/README.md
```diff
@@ -96,19 +78,14 @@ of several other components for which it is a dependency.
 10. Quantifier: <2G, requires Part-of-Speech.
 
 
-### 1.3 LICENSE
```
danyaljj: Moved it to the end of the file.

pipeline/src/main/java/edu/illinois/cs/cogcomp/pipeline/main/PipelineFactory.java
```diff
@@ -40,6 +43,65 @@
...
+                        logger.warn("View name is not supported yet. Look into the readme of the pipeline to see the list of valid annotators. ");
+                }
+            }
+            else {
+                throw new IllegalArgumentException("The view name " + vu + " is not a valid view name. " +
+                        "The possible view names are static members of the class `ViewName`. ");
+            }
+        }
+        // using the default settings and changing the views
+        ResourceManager fullRm = (new PipelineConfigurator()).getConfig(nonDefaultValues);
```
mssammon: Probably need to change defaults in PipelineConfigurator to 'false' for all annotators; corresponding changes in previous PipelineFactory builder to set the flags to true.

danyaljj: So the default would be false for everything. Right?

danyaljj: done.

mssammon: OK, but the default constructor now sets no active views, so pipeline just tokenizes.  Is this desired?

danyaljj: I think that's fine as long as we're clear about it. @bhargav ?

bhargav: Fine for me. Not sure if it would affect other. Maybe we could update the versions to `3.1.x` after this.

mssammon: OK -- please add a comment in README that this is the behavior if default constructor is used. 
Am OK with updating the version as Bhargav suggested.

danyaljj: both done.

danyaljj:
Fixed the issues in pipeline's test.

PR #413, opened 2017-04-05
Fix evalution issue with PredicateArgumentEvaluator.

bhargav:
- `evaluateSense` method requires attributes as part of the  evaluation.
- `evaluate` methods remotes attributes in the gold and predicted views.
- Calling `evaluate` before `evaluateSense` causes the evaluation to  fail as the attributes are missing now.


danyaljj:
Nice catch. I think I fixed it in #215 but it's not checked in.

danyaljj:
Feel free to merge.

bhargav:
Its not fixed in #215. I ran with the changes in that branch and the issue still happens.

danyaljj:
Then I guess I forgot to commit the change.

PR #416, opened 2017-04-09
Removing LBJava files from Quantifier

christos-c:
As per our `.gitignore` file, they should not be tracked


danyaljj:
it seems that files are not being generated upon compile. If we're dropping them, we should set to generate them upon compile.

christos-c:
Oh huh, that's strange. In that case maybe just edit the local `.gitignore` file?

PR #419, opened 2017-04-10
 External annotators 

danyaljj:
Per conversation in email, this is a module which contains external annotators with a pipeline-like class. It also has its own webserver. 
As the first step I added @microth 's PathLSTM SRL annotator, with models deployed with minio. 

Side note: So I did some rebasing which messed up the populated some garbage commits here. Feel free to squash it upon merge.


pom.xml
```diff
@@ -25,8 +25,8 @@
         <module>quantifier</module>
         <module>depparse</module>
         <module>prepsrl</module>
-        <!--pipeline should always be the last item to compile, since some of the previous dependencies have direct dependency on the pipeline-->
```
danyaljj: I think maven compiles them in their right order, no matter what order you put here.

danyaljj:
If your concern is just the download, yeah it will be downloaded and it will be a one-time thing. 
And since it is a separate module, it shouldn't have any other effects. 

The reason that it's here is simple: it's using out datastructures and the webserver from the pipeline. It will be easier to maintain it this way.

PR #435, opened 2017-04-21
fix LineIO file input and output

### LINKED ISSUE #430: LineIO doesn't always work as intended
I had a problem where `LineIO::slurp()` generated an empty string even though I verified the file it was reading was in place and non-empty. This happened ONLY from the **command line**.  Within Intellij, everything worked just fine; also, a test using the same class and same call passed when running `mvn install`.  It was only when I tried running the class via a shell script that the problem arose; writing to STDERR, I could see the file path etc. was correct, and there were no failures -- just an empty string created from the contents of the file.  I solved this by directly using a Scanner/load method in the affected class.
I don't know what would cause the observed behavior -- why there wasn't some kind of exception.  The fact that it did *not* happen in the IDE or mvn install is troubling (the script used the dependencies in target/dependency after `mvn dependency:copy-dependencies`). 
I am reporting this just for the record: I wasted quite a bit of time trying to figure out what was going wrong.

mssammon:
Fixes #430 
text involving characters outside some default range (possibly, characters with diacritics, non-standard quotation marks, etc.) will now be read/written, rather than either a) yielding an empty input or b) substituting some "unk"-type char token for the originals.


core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/io/caches/TextAnnotationMapDBHandler.java
```diff
@@ -30,17 +33,28 @@
...
     public TextAnnotationMapDBHandler(String dbFile) {
-        db = DBMaker.fileDB(dbFile).closeOnJvmShutdown().make();
+        try {
+            db = DBMaker.fileDB(dbFile).closeOnJvmShutdown().make();
+        }
+        catch (DBException e) {
+            logger.warn("mapdb couldn't instantiate db using file '{}': check error and either remove lock, " +
+                    "repair file, or delete file.", dbFile);
+            e.printStackTrace();
+            System.err.println("mapdb couldn't instantiate db using file '" + dbFile +
```
danyaljj: We keeping both `System.err.println` and `logger.warn`?

external/pom.xml
```diff
@@ -69,6 +69,37 @@
...
+                                    <!--<includes>-->
+                                        <!--<include>edu.stanford</include>-->
+                                    <!--</includes>-->
+                                <!--</relocation>-->
+                            <!--</relocations>-->
+                        <!--</configuration>-->
+                    <!--</execution>-->
+                </executions>
+            </plugin>
+        </plugins>
```
danyaljj: do we actually use this?

external/src/test/java/edu/illinois/cs/cogcomp/pipeline/ConflictingDependencyTest.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Properties;
+
+import static junit.framework.TestCase.assertTrue;
+import static org.junit.Assert.fail;
+
+/**
+ * Test combination of regular pipeline and external annotators, which have conflicting versions of Stanford
+ *    libraries. See if shade plugin resolves the problem.
```
danyaljj: Does it?

core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/io/caches/TextAnnotationMapDBHandler.java
mssammon: we could drop System.err.println()...

external/pom.xml
mssammon: not yet. Accidental inclusion.

external/src/test/java/edu/illinois/cs/cogcomp/pipeline/ConflictingDependencyTest.java
mssammon: Hmmm. I thought this was on another branch. No, not yet.

core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/io/caches/TextAnnotationMapDBHandler.java
mssammon: dropped logger.warn() instead; I think of this as cause to exit at present (hence stderr) but it's possible a client could instead delete the offending directory and restart (in which case they can log).  If the mapdb cache is large I assume they may want to try other solutions first.

PR #437, opened 2017-04-24
Upgraded Lucene in Big-Data-Utils

### LINKED ISSUE #239: Upgrade the dependencies of big-data-utils
Upgrade the dependencies of big-data-utils which is introduced in #119 : 

Quote from Mark: 

> This module uses a very old -- and unsupported -- version of mapdb; http://www.mapdb.org/doc/ and Lucene: https://lucene.apache.org/. It may be worth updating to newer versions

shyamupa:
Now works with Lucene 6.0.0. 

Closes #239


big-data-utils/src/main/java/edu/illinois/cs/cogcomp/bigdata/lucene/Lucene.java
```diff
@@ -267,9 +266,9 @@ public static QueryParser newQueryParser(String string) {
 	 * @param id
 	 * @return
 	 */
-	public static Query intQuery(String field, int id) {
-		return NumericRangeQuery.newIntRange(field, id, id, true, true);
-	}
+//	public static Query intQuery(String field, int id) {
```
mssammon: Is this no longer needed?

tokenizer/pom.xml
```diff
@@ -15,17 +15,17 @@
         <dependency>
             <groupId>edu.illinois.cs.cogcomp</groupId>
             <artifactId>illinois-core-utilities</artifactId>
-            <version>3.1.12</version>
+            <version>3.1.13</version>
         </dependency>
         <dependency>
             <groupId>edu.illinois.cs.cogcomp</groupId>
             <artifactId>LBJava-NLP-tools</artifactId>
-            <version>3.1.12</version>
```
mssammon: not sure why all these pom version changes show up; in general, we leave the version as is, and it gets incremented by a deploy build operation on semaphore/jenkins.  Please reset to whatever the current version is.

shyamupa: Ah, I used the increment pom script. Will revert this.

big-data-utils/src/main/java/edu/illinois/cs/cogcomp/bigdata/lucene/Lucene.java
shyamupa: Lucene has removed this kind of queries from the API. All queries are to be made using the QueryParser now.

danyaljj:
@shyamupa could you resolve the conflicts and merge it?

PR #438, opened 2017-04-26
no sentence-pipeline by default in the external.

danyaljj:
Because we have coref annotators, which goes beyond sentences. 

For future: ideally sentence-by-sentence annotations should be set in the `View`-level, rather than `TextAnnotation`-level.

### DIFF
--- a/external/src/main/java/edu/illinois/cs/cogcomp/pipeline/common/ExternalToolsConfigurator.java
+++ b/external/src/main/java/edu/illinois/cs/cogcomp/pipeline/common/ExternalToolsConfigurator.java
@@ -32,7 +32,7 @@ public class ExternalToolsConfigurator extends AnnotatorConfigurator {
      * individually, then splice the annotations together. This allows for partial annotation of
      * documents in cases where document text causes local problems for individual annotators.
      */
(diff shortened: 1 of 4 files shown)


external/src/main/java/edu/illinois/cs/cogcomp/pipeline/common/ExternalToolsConfigurator.java
```diff
@@ -32,7 +32,7 @@
      * individually, then splice the annotations together. This allows for partial annotation of
      * documents in cases where document text causes local problems for individual annotators.
      */
-    public static final Property USE_SENTENCE_PIPELINE = new Property("useSentencePipeline", FALSE);
+    public static final Property USE_SENTENCE_PIPELINE = new Property("useSentencePipeline", TRUE);
```
mssammon: doesn't this set the sentence-based pipeline by default? based on the pr name and the description, this seems like the opposite of what you want.

danyaljj: Ah shoot. You're right; the order in the factory file was reverse. Fixed it now.

PR #455, opened 2017-05-09
adding true-case handler

danyaljj:


external/src/main/java/edu/illinois/cs/cogcomp/pipeline/handlers/StanfordTrueCaseHandler.java
```diff
@@ -0,0 +1,68 @@
...
+        Annotation document = new Annotation(ta.text);
+        pipeline.annotate(document);
+        SpanLabelView vu = new SpanLabelView(viewName, ta);
+
+        String outputString = "";
+
+        for (CoreMap sentence : document.get(CoreAnnotations.SentencesAnnotation.class)) {
+            for (CoreLabel token : sentence.get(CoreAnnotations.TokensAnnotation.class)) {
+                String truecase = token.get(CoreAnnotations.TrueCaseTextAnnotation.class);
+                outputString += " " + truecase;
```
bhargav: If the annotator annotates each token, shouldn't the output be a TokenLabelView which a constituent corresponding to each token in source TextAnnotation. I may have misunderstood how the TrueCase annotator works, just raising a concern.

danyaljj: Changed spanLabelView to TokenLabelView and added the labels per token.

PR #460, opened 2017-05-16
Add test and change privacy of method in CoNLL NER Reader

mayhewsw:
I added the extra test and test file some time ago, and forgot to make a pull request. It tests to make sure that the NER reader reads the right number of sentences. 

Also changed the privacy modifier in CoNLL NER reader so you can read individual CoNLL files instead of whole folders.


danyaljj:
Nice nice nice! 
Merging merging merging!

danyaljj:
No, but with making it static, the functionality (except a little syntax) shouldn't change much. Right?

mssammon:
Right....

PR #461, opened 2017-05-17
Verb sense

### LINKED ISSUE #103: Move verbsense classifier here

danyaljj:
Closes #103

Next PRs: add it to the pipeline.


verbsense/src/main/java/edu/illinois/cs/cogcomp/verbsense/data/PropbankReader.java
```diff
@@ -0,0 +1,116 @@
+/**
```
mssammon: There is a class of the same name in corpusreaders -- if this is duplicate, use the original; otherwise, rename

danyaljj: done.

danyaljj:
r

PR #462, opened 2017-05-20
Removed the hard exception in Constrained PrepSRL

christos-c:
Fixes #458 
Replacing with an error message and returning the local classifier prediction
Minor: added a logger instead of the `System.out` and `System.err` output
Minor: Updated the version of CommaSRL in the pipeline

**NOTE**: Cannot test this locally because of errors while downloading the Maven dependencies (e.g. datastore)

PR #465, opened 2017-05-25
Small fixes for #462

christos-c:
Added an `AssertionError` catcher
Switched PrepSRL view to a `SpanLabelView` since multi-word preposition spans are allowed (also allowing overlapping annotations).


danyaljj:
Jenkins is happy. Merging!

PR #469, opened 2017-05-29
Verb sense classifier to pipeline

danyaljj:


core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/datastructures/ViewNames.java
```diff
@@ -70,6 +70,8 @@
     public static final String SRL_PREP = "SRL_PREP";
     public static final String SRL_COMMA = "SRL_COMMA";
 
+    public static final String VERB_SENSE = "VERB_SENSE";
```
mssammon: you also need to add to the method that indicates the view type -- also in ViewNames.java

danyaljj: done

PR #292, opened 2016-11-26
WIP: adding the scip solver hook and tests.

danyaljj:
Adds the ScipSolver hook + tests. 

Something I'm a little unhappy about is the existence of the binary files in the `lib` folder. 
Any ideas how to get rid of them? 

```
ip-10-0-0-13:lib daniel$ ls -lha
total 33768
drwxr-xr-x  14 daniel  staff   476B Feb  8  2016 .
drwxr-xr-x  15 daniel  staff   510B Nov 26 17:26 ..
-rwxr-xr-x   1 daniel  staff   528K Feb  8  2016 libjscip-0.1.darwin.x86_64.gnu.opt.spx.dylib
lrwxr-xr-x   1 daniel  staff    44B Feb  8  2016 libjscip-0.1.linux.x86_64.gnu.opt.spx.dylib -> libjscip-0.1.darwin.x86_64.gnu.opt.spx.dylib
-rwxr-xr-x   1 daniel  staff   728K Feb  8  2016 libjscip-0.1.linux.x86_64.gnu.opt.spx.so
-rw-r--r--   1 daniel  staff   197K Feb  8  2016 liblpispx-3.1.1.darwin.x86_64.gnu.opt.so
-rw-r--r--   1 daniel  staff   228K Feb  8  2016 liblpispx.linux.x86_64.gnu.opt.so
-rw-r--r--   1 daniel  staff   1.1M Feb  8  2016 libnlpi.cppad-3.1.1.darwin.x86_64.gnu.opt.so
-rw-r--r--   1 daniel  staff   996K Feb  8  2016 libnlpi.cppad.linux.x86_64.gnu.opt.so
-rw-r--r--   1 daniel  staff   4.6M Feb  8  2016 libscip-3.1.1.darwin.x86_64.gnu.opt.so
-rw-r--r--   1 daniel  staff   5.2M Feb  8  2016 libscip.linux.x86_64.gnu.opt.so
-rw-r--r--   1 daniel  staff   1.3M Feb  8  2016 libsoplex-2.0.1.darwin.x86_64.gnu.opt.so
-rw-r--r--   1 daniel  staff   1.5M Feb  8  2016 libsoplex.linux.x86_64.gnu.opt.so
-rw-r--r--   1 daniel  staff   152K Feb  8  2016 scip.jar
```

I'm totally fine to check them in, but I prefer if we find an alternative.

PR #470, opened 2017-05-31
Scip solver

danyaljj:
Same as #292


bhargav:
Relevant: https://github.com/scijava/native-lib-loader

That package can load native libraries from the jar files during runtime. Not sure how much performance degradation happens due to unzipping.

Issue #468, opened 2017-05-28
move external resources as submodules  (instead of modules)


danyaljj:
Currently a few resources are included as single module `external` in addition to `stanford_3.3.1`: 
```
├── external
│   ├── pom.xml
├── stanford_3.3.1
│   ├── pom.xml
```
These two are not merged as one single module, since 

I wanted to add a few more annotators, but it's gonna clutter our project structure if we have too many of these non-Cogcomp modules. Here is my suggested modifications to the current structure. Have a **folder** `external` (rather than a module) which contains submodules for each annotator: 
```
├── external
├──── path-lstm
│   ├──── pom.xml
├──── stanford_3.7.0
│   ├──── pom.xml
├──── stanford_3.3.1
│   ├──── pom.xml
├──── rst-parser
│   ├──── pom.xml
```

what do you think @mssammon ?

mssammon:
This makes sense to me.  I'm even thinking we should have it be a separate project -- full build is now very unwieldy.  What are the reasons to keep it part of cogcomp-nlp?

danyaljj:
About speed, I think we can setup to run additional agents to run things in parallel. 
Having things in one place makes changes easier. 
Alternatively we can intentionally disable tests that are not actively in use (and don't have much importance) (or randomize them: like work in 30% of the times). 

BTW, teamcity gives a good sorting of the tests according to their speed: 
http://morgoth.cs.illinois.edu:5800/viewLog.html?buildId=95&buildTypeId=CogcompNlp_Build&tab=testsInfo

PR #475, opened 2017-06-07
Clausie annotator

danyaljj:
This adds [ClausIE](https://gate.d5.mpi-inf.mpg.de/ClausIEGate/ClausIEGate?inputtext=Bell%2C+a+telecommunication+company%2C+which+is+based+in+Los+Angeles%2C+makes+and+distributes+electronic%2C+computer+and+building+products.&processCcAllVerbs=true&type=true&go=Extract) annotator. 

It's added as a module as it uses a *very* old version of stanford-parser. 
As a future plan, I think it's better to gather external annotators in a folder to keep them separate from our own tools (suggested in #468).


danyaljj:
Added a readme describing what it is and what is does.

PR #477, opened 2017-06-11
updated ner models and fixed XmlTextAnnotationReader bug

mssammon:
also updated ner readme with some info about models including f1 stats on different corpora


mssammon:
The CI failures come from a bug I introduced in an earlier fix for parsing xml corpus annotations. I added a fix to this PR.

mssammon:
@cowchipkid  please take a look as this PR focuses on NER and the corpus reader behavior.

mssammon:
@cowchipkid this is now WIP -- I need to rename the new models, redeploy, etc.

Issue #443, opened 2017-04-28
annotation cache problem after failure


mssammon:
anecdotally, sometimes pipeline goes into infinite loop after annotator failure; presumably, it subtly corrupts the cache? 
after failure, remove corresponding entry from mapdb cache?

mssammon:
see also #391

danyaljj:
This should be resolved after adding transaction. Closing for now. . .

PR #486, opened 2017-06-24
PathLSTM SRL handler minor 

danyaljj:
It hangs on long paragraphs; hence we feed it sentence by sentence.


danyaljj:
In fact I think this is the likely reason for #443

Issue #487, opened 2017-06-24
MapDB: where to do commit 


danyaljj:
In `TextAnnotationMapDBHandler`  don't we need to commit after each [`put`](https://github.com/CogComp/cogcomp-nlp/blob/0beedbdc8a9fa150b2eafa446460340007d31168/core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/io/caches/TextAnnotationMapDBHandler.java#L77) and [`replace`](https://github.com/CogComp/cogcomp-nlp/blob/0beedbdc8a9fa150b2eafa446460340007d31168/core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/io/caches/TextAnnotationMapDBHandler.java#L88)? 
It might slow down the system, but at least upon program crash, the db file won't be empty. Am I right @mssammon ?

danyaljj:
In fact I think this is the likely reason for #443

mssammon:
@danyaljj I think you are right -- I think if someone activates the cache they have to live with it being a little slower in the name of higher reliability.

danyaljj:
Added the fix to #488.

PR #488, opened 2017-06-25
DB commit to TA-DB-Handler

danyaljj:
Fixing #487


danyaljj:
Assuming that @mssammon approved this change in his comment in #487, I'm going to change. If there is any issue we can address later.

Issue #414, opened 2017-04-07
[sentence splitter] Sentence parser stop at incorrect positions 


Slash0BZ:
I found that the sentence parser stops at the second dot in strings like "u.s.", "u.n.", "u.k." and "dr. " causing one sentence to be incorrectly interpreted as two. This raises issues during relation extraction since a gold mention can now cover two sentences, theoretically causing a mention head word to be positioned at a different sentence

Specific examples are:

```
coalition field commanders say u.s .troops are now within 15 miles of the iraqi capital .

u.s .troops could be in baghdad southern outskirts by early thursday .

u.s .officials say they are n't sure whether the iraqi president is still alive .

coming up on our next hour , u.s .forces edged toward the outskirts of baghdad and we 'll tell you what kind of resistance they 're fating from saddam hussein's elite republican guard .

u.s .forces moving in on the airports at saddam hussein international airport , 12 miles west of the city 's center .

bombardment of baghdad continues tonight and over the past 20 minutes , explosions rocking baghdad , central baghdad southern baghdad , the outskirts of baghdad , and importantly , as the coalition forces , principally u.s .army forces , move against saddam hussein international airport -- which is about ten miles outside the center of baghdad -- those explosions are intensifies .

centcom also says 2500 republican guard troops have surrendered to u.s .marines between kut and baghdad .

apparently three ba talians of the u.s .3rd division in baghdad three coalition assets , three huge battle groups in basra itself which the u.k .

troops from the u.s .army 's 101st airborne division went to the site on friday , finding a number of large drums buried in bunkers .

general frankly says it is hard to believe u.n .inspectors would have found the facility hidden as deeply as it was near the euphrates river .

u.s .marines entered southeastern baghdad today and found dozens of chemical suits masks and atropine injecters .

7:00 a.m. , u.s .central command says the coalition may never know if iraqi president saddam hussein survived a u.s .

a u.s .tank shelled the hotel in response to sniper fire .

ejected landed in the desert south of baghdad and then he was rescued by u.s .army ground troops who were in the area .

the u.s .attorney says padilla is an enemy combattant and not entitled to counsel .

very close to the u.s .marines uniform .

lara hijacking in 1995 has been captured by u.s .forces in or near baghdad .

now look as if this man abu absaas in u.s .hands .

u.s .forces in baghdad are staging spot raids around that city , searching for holdout members of saddam 's regime .

[...]

PR #491, opened 2017-06-28
fix tokenizer dot issue

Slash0BZ:
This line fixes the logic where the tokenizer stops in the middle of a word. 

Take "U.S." as an example, when the old version of tokenizer meets the first ".", it will check the next character which is "S". "S" is an alphabet but it is an uppercase character,  which causes the process() fails to return and thus causing the current word to end at the first dot, which it is not supposed to.


danyaljj:
For completeness, it closes #414

danyaljj:
@cowchipkid Dan suggested we train/test a system with and without this change. Wondering if it's possible for you to test NER before/after this change?

cowchipkid:
@danyaljj I will run the usual datasets against this see what we get.

mssammon:
Currently, awaiting check with ACE corpus (which uses offset annotations) instead of NER (which uses column format, meaning that tokenization has no effect)

PR #496, opened 2017-07-01
fix path-lstm.

danyaljj:
When there are redundant constituents in the predicate-argument view, things get messed up upon serialization-deserialization (and it's not noticed until you actually do deserialize it). 

I changed path-lstm handler so that we don't add redundant constituents.

### DIFF
--- a/core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/datastructures/textannotation/View.java
+++ b/core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/datastructures/textannotation/View.java
@@ -416,7 +416,6 @@ public class View implements Serializable, IQueryable<Constituent> {
     public List<Constituent> getConstituentsWithSpan(IntPair span) {
         List<Constituent> list = new ArrayList<>();
         for(Constituent c : this.constituents) if (c.getSpan().equals(span)) list.add(c);
-        list.sort(TextAnnotationUtilities.constituentStartComparator);
         return list;
     }
 
--- a/core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/datastructures/textannotation/View.java
+++ b/core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/datastructures/textannotation/View.java
@@ -8,6 +8,7 @@
 package edu.illinois.cs.cogcomp.core.datastructures.textannotation;
 
 import edu.illinois.cs.cogcomp.core.datastructures.IQueryable;
+import edu.illinois.cs.cogcomp.core.datastructures.IntPair;
 import edu.illinois.cs.cogcomp.core.datastructures.QueryableList;
 import edu.illinois.cs.cogcomp.core.datastructures.ViewNames;
 import edu.illinois.cs.cogcomp.core.transformers.ITransformer;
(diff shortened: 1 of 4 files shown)


external/src/main/java/edu/illinois/cs/cogcomp/pipeline/handlers/PathLSTMHandler.java
```diff
@@ -117,20 +117,20 @@ private PredicateArgumentView getSRL(TextAnnotation ta) throws Exception {
...
                     String label = p.getArgumentTag(a);
                     Yield y = a.getYield(p, label, singleton);
                     IntPair span = new IntPair(y.first().getIdx() - 1, y.last().getIdx());
 
                     assert span.getFirst() <= span.getSecond() : ta;
-                    args.add(new Constituent(label, viewName, ta, span.getFirst(), span.getSecond()));
+                    if(pav.getConstituentsCoveringSpan(span.getFirst(), span.getSecond()).size() == 0)
+                        args.add(new Constituent("Argument", viewName, ta, span.getFirst(), span.getSecond()));
+                    else
+                        args.add(pav.getConstituentsCoveringSpan(span.getFirst(), span.getSecond()).get(0));
```
danyaljj: here is the main change

external/src/main/java/edu/illinois/cs/cogcomp/pipeline/handlers/PathLSTMHandler.java
```diff
@@ -117,20 +117,20 @@ private PredicateArgumentView getSRL(TextAnnotation ta) throws Exception {
...
                 for (Word a : p.getArgMap().keySet()) {
-
                     Set<Word> singleton = new TreeSet<Word>();
                     String label = p.getArgumentTag(a);
                     Yield y = a.getYield(p, label, singleton);
                     IntPair span = new IntPair(y.first().getIdx() - 1, y.last().getIdx());
 
                     assert span.getFirst() <= span.getSecond() : ta;
-                    args.add(new Constituent(label, viewName, ta, span.getFirst(), span.getSecond()));
+                    if(pav.getConstituentsCoveringSpan(span.getFirst(), span.getSecond()).size() == 0)
```
mssammon: This seems like incorrect behavior: shouldn't you confirm that one constituent in the returned list has the same start and end offsets as the one you want to add? it seems like this will incorrectly skip many constituents.

danyaljj: you're absolutely right. See if it looks better now?

core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/datastructures/textannotation/View.java
```diff
@@ -409,6 +410,16 @@ private View createRestrictedView(ITransformer<View, Double> scoreTransformer) {
         return list;
     }
 
+    /**
+     * @return all the constituents that have exact span start/end.
+     */
+    public List<Constituent> getConstituentsWithSpan(IntPair span) {
+        List<Constituent> list = new ArrayList<>();
+        for(Constituent c : this.constituents) if (c.getSpan().equals(span)) list.add(c);
+        list.sort(TextAnnotationUtilities.constituentStartComparator);
```
bhargav: Minor confusion: If the spans are the same, sort should not have any effect right? Might be missing something here.

danyaljj: You're right. Dropped it.

danyaljj:
Merging on green.

PR #497, opened 2017-07-02
Caching changes + minor changes here and there. 

### LINKED ISSUE #450: PipelineConfigurator should extend?
`PipelineConfigurator` should extend `AnnotatorServiceConfigurator`, not `AnnotatorConfigurator`; right?

danyaljj:
- Updated mapdb version. Should fix some cache corruption issues, as discussed here: jankotek/mapdb#706. 
 - Added per-sentence timeout for path-lstm annotator (set to 2mins), because it hangs with long sentences. 
 - Change some "Cogcomp" mentions to "CogComp". 
 - Closes #450

### DIFF
--- a/curator/src/main/java/edu/illinois/cs/cogcomp/curator/CuratorAnnotatorService.java
+++ b/curator/src/main/java/edu/illinois/cs/cogcomp/curator/CuratorAnnotatorService.java
@@ -238,8 +238,11 @@ public class CuratorAnnotatorService implements AnnotatorService {
             return false;
 
         if (annotationCache != null && annotationCache.contains(ta)) {
-            ta = annotationCache.getTextAnnotation(ta);
-            if(ta.getAvailableViews().contains(viewName)) return false;
+            TextAnnotation taFromCache = annotationCache.getTextAnnotation(ta);
+            if(taFromCache.getAvailableViews().contains(viewName)) {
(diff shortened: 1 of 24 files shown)


curator/src/main/java/edu/illinois/cs/cogcomp/curator/CuratorAnnotatorService.java
```diff
@@ -161,6 +185,18 @@ public TextAnnotation createAnnotatedTextAnnotation(String corpusId, String text
...
+            ta = annotationCache.getTextAnnotation(ta);
+            // clone the views
+            HashSet<String> views = new HashSet<>(viewProviders.keySet());
+            views.removeAll(ta.getAvailableViews());
+            if(views.size() == 0) {
+                // it contains the required views; otherwise continue
+                return ta;
+            }
+        }
+
```
danyaljj: Yeah, currently in master, the only place we have caching is inside `addView`. 
Suppose you want to add a bunch of views (like pos, ner, parse, srl). Even if you have it in the cache, you'd have to call this functions for each of these required views. By checking the cache in an earlier stage, I prevent this behavior: if the item is in the cache, we never call `addView` functions and get the result directly. 

If our caching was in the level of Views, I'd totally be for keeping it as it is (i.e. caching in `addView`). But currently caching is done in the level of `TextAnnotation`.

PR #501, opened 2017-07-14
add to utilities for using StringTransformation

mssammon:
function to map TextAnnotation using transformed text to original text offsets
added test
incremented version number


danyaljj:
Looks good to me. 
One question though: do we have a place where we define valid string-transformations? (say for someone who doesn't know how to use it, or when to use it).

mssammon:
Good point.  I've added some documentation to core-utilities' README -- is this sufficient for now?

PR #505, opened 2017-07-16
moving around external annotators: putting everything in a folder cal…

### LINKED ISSUE #468: move external resources as submodules  (instead of modules)

danyaljj:
Changes to external annotators: Following the conversation in #468, 
 - Everything is moved inside a folder "external", so that they don't clutter our standard libraries. All the external annotators will be sub-modules of this folder. 
 - A module named "common" created to prevent repetition. Also, they no-longer depend on the pipeline or core-utils. 
 - Previously-named module "external" is separated into two sub-modules "stanford_380" and "path-lstm". 

Closes #468.


external/external-commons/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+        <artifactId>illinois-cogcomp-nlp</artifactId>
+        <groupId>edu.illinois.cs.cogcomp</groupId>
+        <version>3.1.25</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>external-commons</artifactId>
+    <name>Eexternal Annotators Commons</name>
```
mssammon: typo 'Eexternal'

external/external-commons/src/main/java/edu/illinois/cs/cogcomp/pipeline/server/ExternalAnnotatorsServer.java
```diff
@@ -0,0 +1,223 @@
...
+import java.io.IOException;
+import java.io.UnsupportedEncodingException;
+import java.net.URLDecoder;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import static spark.Spark.*;
+
+public class ExternalAnnotatorsServer {
```
mssammon: Here and elsewhere, please add a Javadoc with basic description

external/external-commons/pom.xml
danyaljj: fixed.

external/external-commons/src/main/java/edu/illinois/cs/cogcomp/pipeline/server/ExternalAnnotatorsServer.java
danyaljj: done.

danyaljj:
Addressed the comments. Any other suggestions?

mssammon:
Looks good to me.

PR #515, opened 2017-07-26
rewrite README for similarity package

lingss0918:


similarity/README.md
```diff
@@ -53,8 +51,25 @@ String s1="please turn on the monitor";
...
 ```
 
-To get the basic LLM similarity score, just set `usePhraseSim` and `useNER` as false in config file (default setting).
+To use this metric properly, you need to specify some configurations in the Config file.
+
+`wordMetric` is the  word comparator metric used in LLM. It can be chosen from "word2vec", "paragram", "esa", "glove", "wordnet" "phrase2vec" or "customized" (your own embedding file).
 
-To use `usePhraseSim` option, set it as true and use `phrase2vec` as `wordMetric`. The system can tokenized the sentence into phrase-based units and it will reformat the sentence. E.g. "please turn the light on" => "please turn-on the light".
+`usePhraseSim` option will automatically tokenized the sentence into phrase-based units when comparing sentences and notice it should be used with "phrase2vec". To use this option, set it as true and use `phrase2vec` as `wordMetric`. The system can identify the phrases and reformat the sentence. E.g. "please turn the light on" => "please turn-on the light". Notice: the phrase identification process depends on the generalized phrases dictionary we extracted from Wordnet (see `src/main/resources/phrases.txt`).
```
mssammon: Please clarify by expanding on this: 
- when you say the system will reformulate the sentence, will it generate an output for the user with the changes, or is it just internal?  If it is just internal, you can rephrase this description to indicate that the system splits the text into phrases, then matches those phrases using a phrase similarity metric that can match different formulations of many phrases, for example 'turn X on' => 'turn on X'. If it is also producing output, say how it can be accessed. (is the input data structure updated?)
Minor points: 
- 'tokenization' refers specifically to identifying word boundaries.  For other levels of representation, you can say "split" or "segment" instead of tokenize.
- here, 'reformulate' is a better description than 'reformat' -- the latter is likely to be interpreted in terms of cleaning up whitespace and maybe punctuation.

similarity/README.md
```diff
@@ -53,8 +51,25 @@ String s1="please turn on the monitor";
...
 
-To use `useNER` option, set it as true. The system will run NER on the sentences first and comparing name entity and words separately. Notice: the NER initialization takes a lot of memory.
+`useNER` option will run NER on sentence and compare name-entity using NE comparison metrics in LLM. To use this option, set it as true. The system will run NER on the sentences first and comparing name entity and words separately. Notice: the NER initialization takes a lot of memory.
```
mssammon: add a pointer to the NER README.

similarity/README.md
```diff
@@ -8,43 +8,41 @@ different similarity resources.
 
 ## Download Resource File
```
danyaljj: Downloading the Resources

similarity/README.md
```diff
@@ -8,43 +8,41 @@ different similarity resources.
...
 
-Once you first use the specific metrics, the system will automatically download corresponding resource file from CogComp server to `user.home` directory in your local machine.
+When you first use the specific metrics, the system will automatically download corresponding resource file from CogComp server to `user.home` directory in your local machine.
```
danyaljj: files

similarity/README.md
```diff
@@ -8,43 +8,41 @@ different similarity resources.
...
 Notice: Some resource file is very large and it may take a while to download. "paragram" is already included in the `src/main/resources/`.
 
 ## Configure File
 
 The default configure file is `config/configurations.properties.` And see the default config in `SimConfigurator` in `edu.illinois.cs.cogcomp.config` package.
```
danyaljj: Why don't we move everything to `SimConfigurator ` and just get rid of `config/configurations.properties`? 

@mssammon  thoughts / preferences on this?

danyaljj: Like ofc it can be an optional setting, but not the default usage.

similarity/README.md
```diff
@@ -8,43 +8,41 @@ different similarity resources.
...
-And the metric can be chosen from "word2vec", "paragram", "esa", "glove", "wordnet" or "phrase2vec" (provided you have downloaded the relevant data resource -- see above).
+
+And the metric can be chosen from "word2vec", "paragram", "esa", "glove", "wordnet", "phrase2vec" or "customized".
+
+In config file, `customized` gives your option to use your own embedding file. Just put the location of the file at this field and the dimension of the embedding at the filed `customized_embedding_dim`.
 
 
 ## Name Entity Comparison
-To use name entity comparison metric:
+When you want to compare Name Entity, you can use name entity comparison metric below:
```
danyaljj: Entities

similarity/README.md
mssammon: I'm fine with reducing the number of files, so OK with just SimConfigurator.

similarity/README.md
danyaljj: Could you clarify this? 
"Notice: the phrase identification process depends on the generalized phrases dictionary we extracted from Wordnet". It is the first time "phrase identification" is mentioned in the text.

lingss0918: updated

similarity/README.md
lingss0918: added

similarity/README.md
lingss0918: done

similarity/README.md
lingss0918: done

similarity/README.md
lingss0918: removed

similarity/README.md
lingss0918: done

similarity/README.md
lingss0918: add explanation

PR #532, opened 2017-08-18
Small additions to external annotators 

danyaljj:
Changes: 
- Often times one of my struggles is to figure out what port is available to run an annotation server.
This scripts automates this process.
- fix a bug in path-lstm: has to switch NomBank and PropBank
- add annotator-specific cache names, so that they don't conflict

### DIFF
--- a/external/external-commons/src/main/java/edu/illinois/cs/cogcomp/pipeline/main/ExternalAnnotatorServiceFactory.java
+++ b/external/external-commons/src/main/java/edu/illinois/cs/cogcomp/pipeline/main/ExternalAnnotatorServiceFactory.java
@@ -29,8 +29,10 @@ public class ExternalAnnotatorServiceFactory {
      * @throws IOException
      * @throws AnnotatorException
      */
-    public static BasicAnnotatorService buildPipeline(Map<String, Annotator> annotators) throws IOException, AnnotatorException {
-        ResourceManager emptyConfig = new ResourceManager(new Properties());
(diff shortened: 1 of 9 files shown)


external/external-commons/src/main/java/edu/illinois/cs/cogcomp/pipeline/main/ExternalAnnotatorServiceFactory.java
```diff
@@ -29,8 +29,10 @@
      * @throws IOException
      * @throws AnnotatorException
      */
-    public static BasicAnnotatorService buildPipeline(Map<String, Annotator> annotators) throws IOException, AnnotatorException {
-        ResourceManager emptyConfig = new ResourceManager(new Properties());
+    public static BasicAnnotatorService buildPipeline(Map<String, Annotator> annotators, String cacheName) throws IOException, AnnotatorException {
```
mssammon: Why change the existing factory method, rather than adding another? we already have many factory methods, so I understand the desire to avoid creating additional ones; but if this parameter is so important, shouldn't other constructors also be modified?

nitpick: choose a name like "cacheDirectory" or "cachePath" rather than "cacheName"

danyaljj: - parameter name: changed. 
- changing the factory name: slightly confused about your comment: are you suggesting adding a another method? Or are you suggesting adding the same parameter to the other constructor? (the answer to which is, the other constructor accepts s ResourceManager parameter, which contains this param).

PR #547, opened 2017-09-10
Closes #546 

### LINKED ISSUE #546: In `TextAnnotationUtilities. copyViewFromTo ` always add the view even if it's empty
Here I think we shouldn't have this if condition. Even if the view is empty, we should add it. 
https://github.com/CogComp/cogcomp-nlp/blob/a39d82de29bb638dff3c2f78a79d6e9250418486/core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/datastructures/textannotation/TextAnnotationUtilities.java#L259

Here is the current issue happening to me. Consider this sentence: 

"It's well."

There is no NERs in this sentence, so NER returns no constituents. With this pattern, we do not populate any NER view (which logically is different from having an empty NER). And I'm running SRL_VERB in the downstream which expects NER view (even if it's empty), but since there no NER view it throws exception. 

@mssammon Do you think it's fine to drop `if` in the line I pointed out?

danyaljj:
- Closes #546 
- Minor name change to the path-lstm to make it compatible with our srl.

### DIFF
--- a/core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/datastructures/textannotation/TextAnnotationUtilities.java
+++ b/core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/datastructures/textannotation/TextAnnotationUtilities.java
@@ -256,15 +256,13 @@ public class TextAnnotationUtilities {
             newVu.addRelation(newR);
         }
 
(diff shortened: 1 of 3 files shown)


danyaljj:
Given the green light by @mssammon in #546, merging this.

PR #549, opened 2017-09-14
Feature Pruning

cowchipkid:
This PR will include the Feature Pruning capabilities.


ner/scripts/benchmarks.sh
```diff
@@ -42,7 +42,7 @@ for JAR in `ls $LIB/*.jar`; do
     cpath="$cpath:$JAR"
 done
 
-CMD="java -classpath  ${cpath} -Xmx12g edu.illinois.cs.cogcomp.ner.NerBenchmark $1 $2 $3 $4 $5"
+CMD="java -classpath  ${cpath} -Xmx14g -Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=8001 -agentpath:/Applications/YourKit-Java-Profiler-2017.02.app/Contents/Resources/bin/mac/libyjpagent.jnilib edu.illinois.cs.cogcomp.ner.NerBenchmark $1 $2 $3 $4 $5"
```
danyaljj: "Xmx12g" -> "Xmx14g"? I thought this change will reduce the memory footprint.

ner/pom.xml
```diff
@@ -61,7 +61,7 @@
         <dependency>
             <groupId>edu.illinois.cs.cogcomp</groupId>
             <artifactId>LBJava</artifactId>
-            <version>1.2.26</version>
+            <version>1.3.0</version>
```
danyaljj: Are we redeploying the models too?

ner/scripts/benchmarks.sh
cowchipkid: @danyaljj That actually depends on several things: is the model they load pruned, is the model one of ours or their own custom model, how many features remain, what gaz do they use, how many examples do they preload. I increased this because I was having failures when loading the unpruned ontonotes. We could reduce this, but it is the max, not the min, so I think it's OK.

ner/pom.xml
cowchipkid: I would really like to do a parameter sweep on ontonotes, more out of curiosity than anything driving need to further improve it. Perhaps we should wait and see if I do that before we rebuild the models? I suspect the sweep on ontonotes only would only take a week or so at the most.

cowchipkid:
Tests are failing still, but not on anything I have changed, just some test fails apparently on a date compare:

[15:58:26] :	 [Step 1/1] Failed tests: 
[15:58:26] :	 [Step 1/1]   Timex3PipelineTest.testSentencePipeline:65 expected:<..."DATE" value="2017-0[7]">> but was:<..."DATE" value="2017-0[8]">>

@danyaljj @mssammon More concerning, The NERAnnotatorTest was failing although it worked on my local machine. During CI "Russian Navy's" results in two NER constituents, Russian and Navy. On my machine, and more accurately, it's one hit, "Russian Navy". If I take the 's off Navy, we don't find the Navy during CI, hence the test passes since we only compare number of constituents. Different Gazetteer somehow??? Different model?

danyaljj:
The temporal package error was supposed to be fixed in #545. If you merge with master, the error will go away.

danyaljj:
> More concerning, The NERAnnotatorTest was failing although it worked on my local machine. During CI "Russian Navy's" results in two NER constituents, Russian and Navy. On my machine, and more accurately, it's one hit, "Russian Navy". If I take the 's off Navy, we don't find the Navy during CI, hence the test passes since we only compare number of constituents. Different Gazetteer somehow??? Different model?

Have you tried testing it on one of our machines? 
Also, I just cleaned all the datastore and maven cache on morgoth. If there is any caching issue, it should go away on CI's next run.

cowchipkid:
@danyaljj According to Git, I

cowchipkid:
This mvn build is a mess. There is WAY too much entirely useless output at build time that one should only see during debugging. And NOBODY should output volumes of text to standard out, it should all be logged.

danyaljj:
Oh that's because `joda-time` get's overriden by something else. It should be fixed after #551. 
But in the meantime, I'd say, pull from the master; if it doesn't go away, add [this](https://github.com/CogComp/cogcomp-nlp/blob/b83654041eaaa932ad6f5643e80c63e6c77500c0/external/external-commons/pom.xml#L31-L35) to your pom file. 

Maven output: Agreed. All that has been accumulated over time. We have people to go through the tests and curb those useless outputs.

cowchipkid:
I just pulled this morning, so I think I am up to date. So that leaves the question of why I was failing on Timex3PipelineTest before, that should have been fixed in 545. Also, why is the NER producing different results on my machine as it is anywhere else. I am going to check two things: the size of the gazetteers file and the models that get loaded. Are these things both loaded by Minio now?

danyaljj:
I'm pretty sure timex3 won't fail anymore, if you pull from the master (you can manually check the git history to make sure https://github.com/CogComp/cogcomp-nlp/pull/545/commits/ec6e87c138f250f23d1b65b3a5884ba5fc46e798  is there). 

Minio: Gazz yes (or at least mostly), models no since they are specified in the pom.xml. 

These are loaded with datastore: 
 - https://github.com/CogComp/cogcomp-nlp/blob/8571a2ef64a5f7a0fd9f243d01bde4d5cdcf60f9/ner/src/main/java/edu/illinois/cs/cogcomp/ner/ExpressiveFeatures/BrownClusters.java#L78-L79 
 - https://github.com/CogComp/cogcomp-nlp/blob/722fc03173e1851c648b1be3dc8b5b9eb8f8589c/ner/src/main/java/edu/illinois/cs/cogcomp/ner/ExpressiveFeatures/TreeGazetteers.java#L63-L64

cowchipkid:
I have looked at the file where the fix was implemented, the fix is not there. When I merge to upstream/master from my fork, it just says I am up to date. Furthermore, the git log shows that commit has been merged. Any suggestions?

danyaljj:
You can ignore timex3 fails. Only make sure NER tests are fine.

cowchipkid:
I could manually replicate that change. But I am wondering what else might have been missed in the merge. I have consistently had problems trying to merge to my fork, it's getting a little frustrating. I think my upstream remote is set up correctly.

cowchipkid:
I have no idea how to close a PR, so I am a little surprised I did. However, this fork is FOOBAR, so I am going to leave this closed til next week and try to get some real work done.

PR #551, opened 2017-09-15
Server client fix

danyaljj:
Changes: 
- Fix server client 
- Upgrade datastore version to the latest version.
- Upgrade gazetteer's version to a newer one, corresponding to bigger ones (that were used to train the system).

### DIFF
--- a/edison/pom.xml
+++ b/edison/pom.xml
@@ -91,7 +91,7 @@
         <dependency>
             <groupId>org.cogcomp</groupId>
             <artifactId>cogcomp-datastore</artifactId>
-            <version>1.9.9</version>
+            <version>1.9.10</version>
         </dependency>
         <!-- Used only in features.FeatureUtilities to convert to LBJava-based feature vectors -->
         <dependency>
--- a/external/external-commons/pom.xml
+++ b/external/external-commons/pom.xml
@@ -26,7 +26,7 @@
         <dependency>
             <groupId>org.cogcomp</groupId>
             <artifactId>cogcomp-datastore</artifactId>
-            <version>1.9.7</version>
+            <version>1.9.10</version>
         </dependency>
         <dependency>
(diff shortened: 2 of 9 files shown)


cowchipkid:
@danyaljj Is there a way to run the tests again without changing something? I would bet if it ran again, it would succeed...

danyaljj:
@cowchipkid yeah it actually went past that, but now it's failing due to memory issues (for which I increased the memory). Any ideas why it was failing in NER though? I feel like it's a recurring theme and we should probably fix it.

cowchipkid:
@danyaljj As far as why failing in NER, the recent one (the seg fault) could have been anything, a latent rare race condition the JVM I suspect, the JVM should never crash because of the code it runs, that's part of the JVM spec. But at the point it crashed in the JVM, it was beyond the NER tests, it had reported the results I think. The previous errors was caused by improved results for the NER (we label "Russian Navy's" entity as one, "Russian Navy" rather than "Russian" and "Navy").

danyaljj:
> @danyaljj As far as why failing in NER, the recent one (the seg fault) could have been anything, a latent rare race condition the JVM I suspect, the JVM should never crash because of the code it runs, that's part of the JVM spec. But at the point it crashed in the JVM, it was beyond the NER tests, it had reported the results I think.  

It might be but it often happens after NER. Any chance that it's correlated to NER?

cowchipkid:
Sure. Maybe memory strain triggers, or some oddness in the NER test code. Loading that MASSIVE gaz does sort of push the JVM. Could be we are hitting some condition that triggers an exit as well, but I wonder where that might be. I didn't see any core files, which is odd, but I am not entirely certain where it would be. And to call system exit and not report something would be... well... par for the course. I could look into this more, but it could be a rabbit hole.

PR #557, opened 2017-09-22
Feature Pruning

cowchipkid:
Feature Pruning is in place, as soon as the Minio repository gets fixed, all will be well.

Same as #549


danyaljj:
@cowchipkid seems like there is a conflict?

chunker/pom.xml
```diff
@@ -19,7 +19,7 @@
         <dependency>
             <groupId>edu.illinois.cs.cogcomp</groupId>
             <artifactId>LBJava</artifactId>
-            <version>1.2.24</version>
+            <version>1.3.0</version>
```
danyaljj: if we're upgrading lbjava version with pruning, ideally we should update all the model files with feature pruning. @mssammon what do you think? 
Btw did e update the parameters after we changed to `double`s to `float`s?

danyaljj: There were some efforts on testing single-precision models but don't remember how things went.

cowchipkid:
@danyaljj I will resolve the conflicts, I suspect it has to do with the merge we just did, in particular, I did change that assert, which I also changed in this branch.

cowchipkid:
@danyaljj @mssammon once again there was either a System.exit or a catastrophic error of some kind. This time, it happened much farther on in the process, during NLP pipeline test. I have looked in the directory it was in, I see no core file, but I don't know that it would produce one. There is an entry in the system log that indicates a segfault, but I have no idea if that has anything to do with this issue or now. What version of Java are we using? How long have we been using team city? Is this a new VM?

danyaljj:
> What version of Java are we using? How long have we been using team city? Is this a new VM? 

We are using jvm8. TC has been working for around 4-5 months.

cowchipkid:
@danyaljj Is that openjdk? Or Sun's Java VM?

danyaljj:
I think it's Sun's java vm.

danyaljj:
Not it's failing because of a memory issue:  http://morgoth.cs.illinois.edu:5800/viewLog.html?buildId=392&

cowchipkid:
Sorry, I have no idea what happened, I must have been looking at the wrong CI entry. Building again now, increased the pipeline to 11g.

cowchipkid:
It's still crashing.

cowchipkid:
I have commits in my fork that are not showing up here. What might cause that?

cowchipkid:
@danyaljj @mssammon I have updated surefire, looks like now it is working. Let's get this merged before something else breaks.

danyaljj:
Looks good to me. TC is happy. Merging.

PR #570, opened 2017-10-19
Minor improvements for NER and MD

Slash0BZ:
support reading brown-clusters from local resource in ner.ExpressiveFeatures.BrownClusters
add a helper function in ColumnFormatReader that transforms TextAnnotation to ColumnFormat

### DIFF
--- a/ner/src/main/java/edu/illinois/cs/cogcomp/ner/ExpressiveFeatures/BrownClusters.java
+++ b/ner/src/main/java/edu/illinois/cs/cogcomp/ner/ExpressiveFeatures/BrownClusters.java
@@ -72,7 +72,7 @@ public class BrownClusters {
      * @param isLowercaseBrownClusters
      */
     public static void init(Vector<String> pathsToClusterFiles, Vector<Integer> thresholds,
-            Vector<Boolean> isLowercaseBrownClusters) {
+            Vector<Boolean> isLowercaseBrownClusters, boolean useLocalBrownCluster) {
 
         try {
         Datastore dsNoCredentials = new Datastore(new ResourceConfigurator().getDefaultConfig());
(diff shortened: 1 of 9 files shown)


ner/src/main/java/edu/illinois/cs/cogcomp/ner/ExpressiveFeatures/BrownClusters.java
```diff
@@ -88,7 +88,14 @@ public static void init(Vector<String> pathsToClusterFiles, Vector<Integer> thre
...
                 // InFile in = new InFile(ResourceUtilities.loadResource(pathsToClusterFiles.elementAt(i)));
-                InputStream is = new FileInputStream(bcDirectory.getPath() + File.separator + pathsToClusterFiles.elementAt(i));
+                // Here we check if the file exists in data-store resources.
+                // If not, we interpret input path as a local resource.
+                String bcFilePath = bcDirectory.getPath() + File.separator + pathsToClusterFiles.elementAt(i);
+                File checker = new File(bcFilePath);
+                if (!checker.exists()){
+                    bcFilePath = pathsToClusterFiles.elementAt(i);
+                }
+                InputStream is = new FileInputStream(bcFilePath);
```
danyaljj: @Slash0BZ this can could confusions. (for people thay may think that it is reading from datastore, but it is in fact reading locally). 
How about your add a switch to the configurator (say `boolean readFromDataStore`, with default value of `true`) and use it here? (rather than checking datastore to see if the file exists or not).

Slash0BZ: Ok that makes more sense.

danyaljj:
Looks good to me. Merging.

PR #572, opened 2017-10-21
WIP: Adding Relation-Extraction package

Slash0BZ:
A extreme light version of https://github.com/Slash0BZ/relation-extraction-experiment, which is experimental code free.


relation-extraction/pom.xml
```diff
@@ -0,0 +1,102 @@
...
+        <maven.compiler.target>1.8</maven.compiler.target>
+    </properties>
+
+    <pluginRepositories>
+        <pluginRepository>
+            <id>CogcompSoftware</id>
+            <name>CogcompSoftware</name>
+            <url>http://cogcomp.cs.illinois.edu/m2repo/</url>
+        </pluginRepository>
+    </pluginRepositories>
```
danyaljj: drop this?

relation-extraction/pom.xml
```diff
@@ -0,0 +1,102 @@
...
+            <id>CogcompSoftware</id>
+            <name>CogcompSoftware</name>
+            <url>http://cogcomp.cs.illinois.edu/m2repo/</url>
+        </repository>
+    </repositories>
+
+    <properties>
+        <maven.compiler.source>1.8</maven.compiler.source>
+        <maven.compiler.target>1.8</maven.compiler.target>
+    </properties>
```
danyaljj: drop this?

relation-extraction/src/main/java/edu/illinois/cs/cogcomp/nlp/corpusreaders/ACEReader.java
```diff
@@ -0,0 +1,558 @@
+/**
```
danyaljj: Isn't this duplicate, or has to be moved to the corpus-readers?

relation-extraction/src/main/java/edu/illinois/cs/cogcomp/pipeline/handlers/StanfordTrueCaseHandler.java
```diff
@@ -0,0 +1,72 @@
+/**
```
danyaljj: This is duplicate; right?

Slash0BZ: I mean yeah, the same one is also included in MD. But this class is a part of another package (neither MD or RE), I am reloading this class here, so I think I have to include it under src/ ? Let me know if there is another way.

relation-extraction/pom.xml
Slash0BZ: Just realized that I can't drop this if we want to generate LBJ class on the fly like NER. Otherwise I need to commit all generated classes as well.

relation-extraction/src/main/java/edu/illinois/cs/cogcomp/pipeline/handlers/StanfordTrueCaseHandler.java
danyaljj: This is using an older version of stanford; right? If so, why not move it to [here](https://github.com/CogComp/cogcomp-nlp/tree/master/external/stanford_3.3.1/src/main/java/edu/illinois/cs/cogcomp/pipeline/handlers)?

danyaljj:
A few suggested changes: 
- Add a readme. 
- Add link to the readme of the root folder. 
- Add your annotator to the pipeline and `PipelineFactory`, and test that your annotator properly works when calling from the pipeline.

relation-extraction/src/main/java/edu/illinois/cs/cogcomp/pipeline/handlers/StanfordTrueCaseHandler.java
Slash0BZ: Sounds good

danyaljj:
Looks good to me.  Merging!

PR #579, opened 2017-10-31
Fix RE annotator exception & Add entityType as label in MD

Slash0BZ:
1) RelationAnnotator now only copy the mentionView for ViewNames.RELATION if no mentions are predicted
2) MentionAnnotator now uses the entity type for each mention as its Constituent label, so that the types show up in the demo

### DIFF
--- a/md/src/main/java/org/cogcomp/md/BIOTester.java
+++ b/md/src/main/java/org/cogcomp/md/BIOTester.java
@@ -344,22 +344,17 @@ public class BIOTester {
             }
         }
         String entityType = goldType;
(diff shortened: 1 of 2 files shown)


Slash0BZ:
I am doing pipeline tests. I will let you know @danyaljj

Slash0BZ:
My local pipeline tests look good. I tested both cases when there are mentions and not.

Slash0BZ:
My local pipeline results look good

danyaljj:
Cool thanks

Issue #580, opened 2017-10-31
duplicate files in md


mssammon:
These are apparently duplicates or replications of files in other modules ('external', 'corpusreaders', and 'tokenizer').  

If they are duplicates, replace them.
If they are modified versions, unify them.

https://github.com/CogComp/cogcomp-nlp/blob/master/md/src/main/java/edu/illinois/cs/cogcomp/pipeline/handlers/StanfordTrueCaseHandler.java

https://github.com/CogComp/cogcomp-nlp/blob/master/md/src/main/java/edu/illinois/cs/cogcomp/nlp/corpusreaders/ACEReader.java

https://github.com/CogComp/cogcomp-nlp/blob/master/md/src/main/java/edu/illinois/cs/cogcomp/nlp/tokenizer/TokenizerStateMachine.java

Slash0BZ:
Ok I will handle it

Slash0BZ:
For ACE reader: I think I have to keep this version unless 
1) we merge the true-cased ACEReader into corpusreader 
2) we discard true-casing in both MD and RE
@danyaljj We don't want the true-cased ACEReader in corpusreader right?

danyaljj:
> We don't want the true-cased ACEReader in corpusreader  

True. is it possible to make it a class that extends the ACEReader of corpusreader and overrides the method that uses the true-caser? 
Also, if you change the name it would reduce the confusion. (For example `ACEReaderWithTrueCaseFixer`).
An alternative method is to write a function that given TextAnnotation with messed up casing, it fixes the casing and creates a new TextAnnotation. This was you can reuse the ACEReader of corpusreaders.

mssammon:
I really like the idea of a truecaser function...

Slash0BZ:
I will first try to extent the class. I think it should work.

PR #581, opened 2017-11-01
add a ACEReaderWithTrueCaseFixer class

Slash0BZ:
update usages upon this new class
Also a bug fix that cases opposite direction relations in RelationAnnotator.
Also added two new testing functions in MD and RE.


Slash0BZ:
closes #580

danyaljj:
I was thinking that it might be a good idea if we move `ACEReaderWithTrueCaseFixer.java` (which is in corpus readers) to `external/stanford_3.3.1` (hence we don't have to add stanford as a dependency to corpus-readers). What do you think? (@mssammon feel free to comment on this)

Slash0BZ:
Ah that way ACEReaderWithTrueCaseFixer will be a package in stanford_3.3.1 so it can't extend ACEReader anymore right?

mssammon:
the new reader can be in the external/ package, I agree... maybe point to it in the corpusreader documentation, though.  I don't see why the reader can't extend ACEReader; package name/path doesn't have to be identical.

danyaljj:
So rather than copus-readers being dependant on stanford, stanford would be dependant on stanford.

Slash0BZ:
Just making sure, are we moving it to `edu.illinois.cs.cogcomp.pipeline.common` which holds all the external sources?

danyaljj:
No. Inside `stanford_3.3.1`: 
https://github.com/CogComp/cogcomp-nlp/tree/master/external/stanford_3.3.1

Slash0BZ:
I understand, but all files in `stanford_3.3.1` are in package `edu.illinois.cs.cogcomp.pipeline.common`.

danyaljj:
Ah you can keep the package naming compatible with corpus reader (which is what you currently have).

external/stanford_3.3.1/pom.xml
```diff
@@ -33,5 +33,10 @@
             <version>3.3.1</version>
             <classifier>models</classifier>
         </dependency>
+        <dependency>
+            <groupId>edu.illinois.cs.cogcomp</groupId>
+            <artifactId>illinois-corpusreaders</artifactId>
+            <version>3.1.35</version>
```
danyaljj: 3.1.36

danyaljj:
Minor comment on versioning: merge after fixing it (and CI turning green).

PR #583, opened 2017-11-02
Single Precision DVector

cowchipkid:
The DVector now stores floats, can read floats or doubles, but only writes floats.


core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/datastructures/textannotation/SpanLabelView.java
```diff
@@ -62,7 +62,17 @@ public SpanLabelView(String viewName, String viewGenerator, TextAnnotation text,
     public void addConstituent(Constituent constituent) {
         super.addConstituent(constituent);
 
-        Collections.sort(this.constituents, TextAnnotationUtilities.constituentStartComparator);
```
mssammon: is it worth simply replacing the map with a TreeMap, as you suggest? if not now, please open an issue.

ner/scripts/paramsweep/generatesweep.sh
```diff
@@ -5,15 +5,151 @@
...
-	    NEWDIR="r"$i"-t"$j
-	    cp -rf ./cogcomp-nlp/ ./$NEWDIR
-	    NEWNERDIR=$NEWDIR"/ner"
-	    sed 's/SparseAveragedPerceptron\([^\)]*\)/SparseAveragedPerceptron\('"$i"', 0, '"$j"'/' $NEWNERDIR/src/main/lbj/LbjTagger.lbj > $NEWNERDIR/src/main/lbj/poogers.lbj
-	    mv $NEWNERDIR/src/main/lbj/poogers.lbj $NEWNERDIR/src/main/lbj/LbjTagger.lbj
-    done;
-done;
\ No newline at end of file
+#
+# first build cogcomp-nlp. Oddly enought, take much long to build than to
```
mssammon: typo 'enought'

mssammon: just to show I'm paying attention :-)

ner/scripts/paramsweep/generatesweep.sh
```diff
@@ -5,15 +5,151 @@
...
+learnrateL1=(.01 .03)
+thicknessL1=(10 30)
+learnrateL2=(.02 .04)
+thicknessL2=(20 40)
+
+#
+# copy the code all over the place.
+#
+echo
+echo -------- cloning code into new directories --------------
```
mssammon: seems like this will take up a lot of space...  we may want to think about how to refine it.   Is it plausible to simply change output paths, or at least just create symlinks -- I'm assuming the run script launches separate processes for each parameter set; even in parallel, would this create any problems?

ner/src/main/java/edu/illinois/cs/cogcomp/ner/BenchmarkOutputParser.java
```diff
@@ -0,0 +1,356 @@
...
+    /**
+     * This main method will take one required argument, idenfitying the file containing 
+     * the results. Optionally, "-single" may also be passed indicating it will extract
+     * the F1 value for single token values only.
+     * @param args
+     * @throws IOException 
+     */
+    public static void main1(String[] args) throws IOException {
+        /** get the F1 score for token only.  */
+        String fuck = "Token-level Acc Level1:\n" +
```
mssammon: While guilty of the odd obscenity myself in comments, maybe rename this variable?

core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/datastructures/textannotation/SpanLabelView.java
cowchipkid: This is an inherited data structure. There is already an issue open for this. I will revise the existing issue.

ner/scripts/paramsweep/generatesweep.sh
cowchipkid: I already know how to refine this, but way out of scope for this release. This is already *much* smaller that it was in the previous iteration. The right way to do this is to run them all in one VM. Makes the hairs on the back of my neck rise. Much smaller memory footprint, much faster, lots less disk.

ner/src/main/java/edu/illinois/cs/cogcomp/ner/BenchmarkOutputParser.java
cowchipkid: hehehe. Whoops. that was a test method and should have been removed.

danyaljj:
Looks good to me. 
@mssammon  ?

mssammon:
Good to merge and deploy.

PR #584, opened 2017-11-02
clean up some old depndencies in transliterator.

danyaljj:
Apparently, some of the transliteration dependencies were bringing some of our old dependencies that are not supported anymore (e.g. an old version of core-utils). Dropping dependencies that we don't use anymore. 

FYI @HornHehhf


mayhewsw:
These changes look fine. If it still compiles, I will accept (I can't see on TeamCity).

mayhewsw:
Oh, I see now. It doesn't compile.

danyaljj:
Fixed the compile issues.

Issue #595, opened 2017-12-25
 IO Error while reading the file at hierarchies/20newsgroups/parentChildIdMap.txt


linxiexiong:
when running ./pipline/scripts/runWebserver.sh to start remote server, it comes out problem "ERROR DatalessAnnotatorUtils:104 - IO Error while reading the file at hierarchies/20newsgroups/parentChildIdMap.txt", while i can't find parentChildIdMap.txt this file.

danyaljj:
Thanks! @shatu is working on this issue I believe. 
In the mean time you can use an older release.

danyaljj:
Should be fixed now.

PR #601, opened 2018-01-24
Question type classification + .... 

danyaljj:
In this PR, we:  

- Add the question-typer as a module: the question typer is based on the data annotated by (Lin&Roth, 2003). 
- Add the question-typer into the pipeline 
- Move the pipeline client to outside pipeline; the rationale is that, one who needs to send calls to the pipeline over a network (say when I am training a classifier) I don't have to import everything in the pipeline (and all the headaches that would come with importing many random dependencies). 
- Make dataless annotator lazy, so that they don't trip the pipeline upon its load (related to #595) FYI @shatu 
- Add an API to pipeline server to make its version available.


danyaljj:
Finally passing the tests! 😄

PR #607, opened 2018-02-16
Update README.md

cowchipkid:
These results are *not* as good as previous results! I don't know why.


cogcomp-dev:
Seems like this is due to a gazetteer bug.  Will await fix and update.

PR #612, opened 2018-02-18
adding datastore models for ner.

danyaljj:
Using datastore to load NER models.

### DIFF
--- a/ner/src/main/java/edu/illinois/cs/cogcomp/ner/NERAnnotator.java
+++ b/ner/src/main/java/edu/illinois/cs/cogcomp/ner/NERAnnotator.java
@@ -8,6 +8,7 @@
 package edu.illinois.cs.cogcomp.ner;
 
 import edu.illinois.cs.cogcomp.annotation.AnnotatorConfigurator;
+import edu.illinois.cs.cogcomp.core.resources.ResourceConfigurator;
 import edu.illinois.cs.cogcomp.core.utilities.configuration.Configurator;
 import edu.illinois.cs.cogcomp.lbjava.learn.Lexicon;
 import edu.illinois.cs.cogcomp.ner.ExpressiveFeatures.ExpressiveFeaturesAnnotator;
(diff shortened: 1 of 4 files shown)


ner/src/main/java/edu/illinois/cs/cogcomp/ner/config/NerBaseConfigurator.java
```diff
@@ -35,6 +35,10 @@
     public final static String PATH_TO_MODEL = "pathToModelFile";
     public final static String VIEW_NAME = "viewName";
 
+    public final static String TRAINED_ON = "all-data";
```
danyaljj: To choose whether we use the models that are train on: 
- all data 
or 
- training data only

ner/src/main/java/edu/illinois/cs/cogcomp/ner/config/NerBaseConfigurator.java
```diff
@@ -130,6 +134,7 @@ public ResourceManager getDefaultConfig() {
         Properties props = new Properties();
 
         props.setProperty(VIEW_NAME, ViewNames.NER_CONLL);
+        props.setProperty(TRAINED_ON, TRAINED_ON_ALL_DATA);
```
danyaljj: the default is "models trained on all data"

ner/src/main/java/edu/illinois/cs/cogcomp/ner/NERAnnotator.java
```diff
@@ -100,20 +106,37 @@ public void initialize(ResourceManager nerRm) {
...
+        else {
+            dataset = "ontonotes";
+        }
+
+        String data_split = nerRm.getString(NerBaseConfigurator.TRAINED_ON);
+
+        try {
+            Datastore ds = new Datastore(new ResourceConfigurator().getDefaultConfig());
+            String artifact_id = "ner-model-" + dataset + "-" + data_split;
+            File modelDir = ds.getDirectory("edu.illinois.cs.cogcomp.ner", artifact_id, 4.0, false);
```
danyaljj: load the model from datastore here.

ner/src/main/java/edu/illinois/cs/cogcomp/ner/NERAnnotator.java
```diff
@@ -100,20 +106,37 @@ public void initialize(ResourceManager nerRm) {
...
+            dataset = "enron-conll";
+        }
+        else {
+            dataset = "ontonotes";
+        }
+
+        String data_split = nerRm.getString(NerBaseConfigurator.TRAINED_ON);
+
+        try {
+            Datastore ds = new Datastore(new ResourceConfigurator().getDefaultConfig());
```
cogcomp-dev: Shouldn't this use the ResourceManager argument provided to the constructor to overwrite any properties the user wants to configure?

ner/src/main/java/edu/illinois/cs/cogcomp/ner/config/NerBaseConfigurator.java
cogcomp-dev: Nice.

ner/src/main/java/edu/illinois/cs/cogcomp/ner/NERAnnotator.java
danyaljj: Good point. Updated it.

PR #614, opened 2018-02-19
Ner models

danyaljj:
drop maven dependencies for NER models.


danyaljj:
Closing this since I think it's already incorporated.

PR #624, opened 2018-03-24
Minor IO changes to the MASC reader

lxtbell:


danyaljj:
Just a heads up: the test was failing since the CuratorFactory Test (which is likely because curator is off).

cogcomp-dev:
I know -- I checked the failure report and saw it was Curator. We just had a security sweep on the servers and that invariably chokes up curator and other demos. The changes are very local and will have no effect outside the MASC reader. (The changes to IOUtils relocates a general snippet from corpusreaders to coreutils.)

PR #625, opened 2018-03-24
Nesim type

sanjayss34:
Allow user to provide types for one or both names in NESim.compare()


danyaljj:
Thanks @sanjayss34 ! 
Two minor comments: 
Could you please update the readme as well? (i.e. examples where we use NESim and how to enter the new arguments, possibly with a few examples).

Issue #610, opened 2018-02-17
Temporal normalization: include type of the normalization 


danyaljj:
Take this example: 
"He was born 10 years ago. I started the run this morning. I have been this mad since April of last year. 14th of February is valentines day. I have been running for 3 hours, today. "

There multiple *types* of normalizations: 

[<TIMEX3 type="DATE" value="2005"> 10 years ago ]   --> normalizing year 
[<TIMEX3 type="TIME" value="2015-01-01TMO"> this morning ]  --> normalizing time of the day 
[<TIMEX3 type="DATE" value="2014-04"> April of last year ] --> normalizing year 
[<TIMEX3 type="DURATION" value="PT3H"> 3 hours ] --> normalizing hour 
[<TIMEX3 type="DATE" value="2015-01-01"> today ] --> normalizing day 

Can we populate this information in the output view? 

Btw: suggestion was made by @snigdhac

danyaljj:
@666666fzl @qiangning thoughts?

666666fzl:
Hi Daniel,

I can't see the previous discussion, can you provide some context? Thanks.

Best regards,
Zhili Feng

danyaljj:
https://github.com/CogComp/cogcomp-nlp/issues/610#issuecomment-366459091

Daniel

666666fzl:
Hi @danyaljj , what do you mean by "populate"?

danyaljj:
By populating I mean, is it possible to extract such information? (and possibly put them inside the temporal `View` of the output `TextAnnotation`)

666666fzl:
I thought we include that already? https://github.com/CogComp/cogcomp-nlp/blob/master/temporal-normalizer/src/main/java/edu/illinois/cs/cogcomp/temporal/normalizer/main/timex2interval/TimexChunk.java,
"type" is part of the "attributes" map?

PR #630, opened 2018-04-15
Populate timex normalization type to the TIMEX View in TextAnnotation

qiangning:
- Modified `TemporalChunkerAnnotator.java` so that the `type` and `value` attributes are populated to the `ViewNames.TIMEX3` view of TextAnnotation.
- Created a `Demo.java` to show how to use TemporalNormalizer as well as how to get the `type` and `value` attributes programmatically.


danyaljj:
Looks good to me! Merging! 

Closes #610

PR #634, opened 2018-04-20
added sample-based stratified dev/train/test splitter

mssammon:
original corpus splitter was designed for ERE corpus, which is small; the evaluation of different splits was very inefficient.  Added a new version that performs sequential splits based on user-specified properties of corpus, preferring to split by infrequent features, and trying several random sample splits at each step. Cost is computed as divergence from "ideal" split, which is based on the frequency of the user-specified properties. The actual split is based on a representation that allows e.g. all examples from a given document to be kept in the same split (avoids potential "memorization" by learner).


mssammon:
corrected error: I initially used an out-of-date branch for this PR.  Now updated; also made some minor improvements to documentation in various utility classes.

Issue #609, opened 2018-02-17
Temporal Normalizer: does get short forms of months. 


danyaljj:
For example, it doesn't get this example: 
"Feb 14 is valentines day."

PR #636, opened 2018-04-22
Fix iss #609: temporal component cannot capture obvious timexes

qiangning:
Issue #609: Temporal Normalizer: does get short forms of months.
For example, it doesn't get this example:
"Feb 14 is valentines day."

This is due to our design. We use a two-step procedure in our normalizer:
1. Use chunker to chunk out time expressions (timex). 
2. Use our normalizer to convert timex-es from surface forms to standard values (e.g., 2018-04-21). The purpose of this design was to reduce the computational time of conventional methods (e.g., HeidelTime). 

However, the cost is that step 1 (chunker) is not perfect and sometimes fails on very easy cases (here "Feb 14" is one example). 

Above is the reason of such behavior. Although it is expected, I think it's worthwhile to at least use some simple rules before we apply the chunker so that we don't make silly mistakes. Specifically, after chunker makes a decision about whether a token is B/I/O, if it's O, then I do a sanity check to see if this token is a month/week-of-day name (or abbr.). If it is indeed such a token, I revert it from O to B.

Now if we feed `-t "Feb. 14 is the Valentines day."` to `edu.illinois.cs.cogcomp.temporal.normalizer.main.Demo`, we get it correctly:
```
There're 1 time expressions (TIMEX) in total.
TIMEX #0: Text=Feb. 14, Type=DATE, Value=2018-02-14
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,8 @@ edison/wordnet/
 **/.project
 **/.classpath
 *~
+
+# lbjava auto-generated files
+question-type/src/main/java/edu/illinois/cs/cogcomp/question_typer/lbjava
+ner/src/main/java/edu/illinois/cs/cogcomp/ner/LbjFeatures
+relation-extraction/src/main/java/org/cogcomp/re/LbjGen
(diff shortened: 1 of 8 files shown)


danyaljj:
Beautiful! Thanks, Qiang.

PR #639, opened 2018-04-24
fix a TextAnnotation builder bug on Windows

Slash0BZ:
On Windows platform, `System.lineSeparator()` is `\r\n` instead of `\n`

This causes issue when creating a TextAnnotation. Every sentence should start with the previous ending character offset plus the size of the lineSeparator, as specified at https://github.com/CogComp/cogcomp-nlp/blob/1267fc72aba945a77249d34787ecb607eaa81f3b/core-utilities/src/main/java/edu/illinois/cs/cogcomp/annotation/BasicTextAnnotationBuilder.java#L56

An alternative fix will be replacing the above line with `text += StringUtils.join(sentenceTokens, ' ') + "\n";`

Please let me know which one is a better fix.

I actually lean towards the second one, but I don't know if there is a reason for line 56 to use `System.lineSeparator()` instead of `\n`.

@mssammon @danyaljj


mssammon:
The idea of using `System.lineSeparator()` is to be platform-neutral, so probably the correct approach is to use the width of `System.lineSeparator()`.  The challenge is to do it consistently...

danyaljj:
@mssammon just to make sure: you are in favor of keeping this PR as it is; correct?

mssammon:
Yes, unless you object, @danyaljj

PR #647, opened 2018-05-16
Tokenizer changes

cowchipkid:
Add option to split on multiple newlines, capture emails as single token. Upped version to 4.0.5


mssammon:
@cowchipkid seeing a test failure in verbsense module -- can you take a look and see if your changes are the cause?

danyaljj:
Yeah the system crashes at verbsense: 
http://morgoth.cs.illinois.edu:5800/viewLog.html?buildId=684&buildTypeId=CogcompNlp_Build&tab=buildLog&filter=err

danyaljj:
Actually, we're getting this error elsewhere too. Might be completely independent of this PR.

PR #649, opened 2018-05-17
disabled problematic MASC reader test; removed edison test output files.

mssammon:
disabled MASC corpus reader test to avoid failure when run on a non-CCG machine.
removed edison test output files, which I don't think should be in the git repo as they are outputs, and are not invariant across tests (at least not on different machines).


danyaljj:
Looks good to me. 

Btw, did you see why verbsense keep failing? I suspect it has something to do with restarting the morgoth server. Maybe it is missing Gurobi server?

Issue #648, opened 2018-05-17
make MASC reader test portable


mssammon:
Currently MASCXCESReaderTest fails if it is run on a non-CCG machine due to a hard-coded path that points outside the project:

https://github.com/CogComp/cogcomp-nlp/blob/master/corpusreaders/src/test/java/edu/illinois/cs/cogcomp/nlp/corpusreaders/mascReaderTests/MascXCESReaderTest.java#L27

Please copy a small subset of MASC XCES files to test/resources/ and point the test there.

mssammon:
resolved by #650

PR #650, opened 2018-05-18
Add MascXCESReaderTest corpus to resources

lxtbell:
Fix MascXCESReaderTest failure when run on a non-CCG machine.


lxtbell:
#648

lxtbell:
These are already fragments but I've just made them even smaller.

Issue #651, opened 2018-05-23
StatefulTokenizer weirdness


borice:
Hello.

TL;DR;
StatefulTokenizer tokenizes the date "10/23/2018" as [ "10", "/", "23/2018" ] whereas IllinoisTokenizer (which seems to be deprecated) keeps it as a single token [ "10/23/2018" ].  

Longer version:
I'm seeing some unexpected sentence detection/tokenization of a simple (contrived) text.
Here's an example (in Scala, but can easily be translated to Java)

   ```
    val textEN = """One two, three-four-five 10/23/2018 at 5:20pm one? Of course not! Be well, stranger. Bye-bye!"""
    val textAnnotationBuilder = new TokenizerTextAnnotationBuilder(new StatefulTokenizer(false))
    val textAnnotation = textAnnotationBuilder.createTextAnnotation(text)
    val sentences = textAnnotation.sentences().asScala.toList
    val sentenceTokens = sentences.map(_.getTokens)

```
Printing the sentences and sentence tokens I get:

> SENTENCES:
> 1. One two , three-four-five 10 / 23/2018 at 5:20pm one ?
> 2. Of course not !
> 3. Be well , stranger .
> 4. Bye-bye !
> 
> SENTENCE TOKENS:
> 1. One | two | , | three-four-five | 10 | / | 23/2018 | at | 5:20pm | one | ?
> 2. Of | course | not | !
> 3. Be | well | , | stranger | .
> 4. Bye-bye | !

**Problem 1**
The problem is in the date "10/23/2018"  -- it has no spaces in the original text, but somehow the sentence output added some spaces "10 / 23/2018".  Each sentence was printed with `Sentence.toString()` method.  If I use the `Sentence.getText()` method instead, the sentence is printed correctly with no extra spaces in the date.

**Problem 2**
This seems related to above. As you can notice, the sentence tokens also seem to split the date in a weird way --- which seems to correspond to the extra spaces issue described above.

If instead of `StatefulTokenizer` I use `IllinoisTokenizer` (now seemingly deprecated), everything works as it should.  Both `Sentence.toString()` and `Sentence.getText()` output the same thing, and the sentence tokens do not have the extra-space problem.

Please advise.

Thank you,
Boris

PR #654, opened 2018-05-24
Tokenizer/NER Model Loading

cowchipkid:
Fixed issue #651 at least in part, common dates like "10/14/2016" are not parsed to a single token. I suspect that will fix part 2 of the issue as well, since the "10" part is not now it's own token. Also change the NER model loader so that the only time the default models are loaded is when the view name is one of the default view names, ViewNames.NER_CONLL or ViewNames.NER_ONTONOTES.


danyaljj:
Looks good to me. Merging.

Issue #660, opened 2018-06-11
error in serialized json output for TextAnnotation


mssammon:
...when there are multiple labels + scores. key/value pairs should be written into a named json element but instead are just written to constituent element:
`https://github.com/CogComp/cogcomp-nlp/blob/master/core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/utilities/JsonSerializer.java#L406`

PR #662, opened 2018-06-11
fixed a bug in json serializer.  Also updated to explicitly store and…

### LINKED ISSUE #661: sentence and token views not serialized, but they should be
We have an explicit sentence view: make sure it is written out when serialized, and that it is read when deserialized.  Check deserialization does not create errors.

mssammon:
… read SENTENCE view. Closes #661  and #660

### DIFF
--- a/core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/utilities/JsonSerializer.java
+++ b/core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/utilities/JsonSerializer.java
@@ -349,7 +349,6 @@ public class JsonSerializer extends AbstractSerializer {
 
     private static String readString(String name, JsonObject obj) {
         return obj.getAsJsonPrimitive(name).getAsString();
-        // return obj.get(name).getAsString();
     }
 
     private static void writeInt(String name, int value, JsonObject out) {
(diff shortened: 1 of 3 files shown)


core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/utilities/JsonSerializer.java
```diff
@@ -435,9 +435,8 @@ JsonObject writeTextAnnotation(TextAnnotation ta, boolean doWriteTokenOffsets) {
 
         JsonArray views = new JsonArray();
         for (String viewName : Sorters.sortSet(ta.getAvailableViews())) {
-            // TODO: comment out this next two lines as part of addressing issue #406
-            if (viewName.equals(ViewNames.SENTENCE))
-                continue;
+//            if (viewName.equals(ViewNames.SENTENCE))
+//                continue;
```
ChaseDuncan: Why are these being left in as comments?

core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/utilities/protobuf/ProtobufSerializer.java
```diff
@@ -220,8 +220,8 @@ public static TextAnnotationProto writeTextAnnotation(TextAnnotation ta) {
         writeSentences(ta, taBuilder);
 
         for (String viewName : Sorters.sortSet(ta.getAvailableViews())) {
-            if (viewName.equals(ViewNames.SENTENCE))
-                continue;
+//            if (viewName.equals(ViewNames.SENTENCE))
+//                continue;
```
ChaseDuncan: Why are these being left in as comments?

mssammon: good catch. Removed and pushed changes.

PR #663, opened 2018-06-14
fix the timex bug when may and sat appear as verbs instead of timexes

qiangning:
Previously, to avoid silly mistakes, I incorporated rules before timex chunker, so that any word matching to month names or week-of-the-day names will be forced to be a timex.

However, I realized that this rule is not accurate. `may` can be either a modal verb, or the month. `sat` can be also the past tence of `sit`. So I have added a POS check now: only those `may`'s or `sat`'s that are NNP will be forced to be timex.

`sun` is a bit tricky, since the star Sun is also NNP. I pick it out of our rules and leave it up to the chunker's decision.


qiangning:
Daniel @danyaljj, Mark @mssammon : I suggest Zhili @666666fzl review this change.

qiangning:
@danyaljj I've merged this PR. Since this is my first time merging PRs, are there anything else I need to do after the merge?

danyaljj:
Nope. You're good. 

Note that the change is not deployed yet. In other words, the jar files in the maven repo (http://cogcomp.org/m2repo/edu/illinois/cs/cogcomp/) still don't contain your changes. We deploy only when there is a need for doing so. If you need it, you have to deploy it (details [here](https://github.com/CogComp/cogcomp-nlp/wiki/Git-pull-requests-and-Maven-deployment#quick-deploy))

PR #667, opened 2018-07-06
Incremental Training

cowchipkid:
This version supports incremental training of models. Use the -incremental argument of NerBenchmark to further training of an existing model.

### DIFF
--- a/ner/src/main/java/edu/illinois/cs/cogcomp/ner/LbjTagger/LearningCurveMultiDataset.java
+++ b/ner/src/main/java/edu/illinois/cs/cogcomp/ner/LbjTagger/LearningCurveMultiDataset.java
@@ -12,7 +12,6 @@ import edu.illinois.cs.cogcomp.lbjava.classify.TestDiscrete;
 import edu.illinois.cs.cogcomp.lbjava.learn.BatchTrainer;
 import edu.illinois.cs.cogcomp.lbjava.learn.SparseAveragedPerceptron;
 import edu.illinois.cs.cogcomp.lbjava.learn.SparseNetworkLearner;
-import edu.illinois.cs.cogcomp.lbjava.learn.featurepruning.SparseNetworkOptimizer;
 import edu.illinois.cs.cogcomp.lbjava.parse.Parser;
 import edu.illinois.cs.cogcomp.ner.ExpressiveFeatures.ExpressiveFeaturesAnnotator;
 import edu.illinois.cs.cogcomp.ner.ExpressiveFeatures.TwoLayerPredictionAggregationFeatures;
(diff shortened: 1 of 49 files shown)


ner/src/main/java/edu/illinois/cs/cogcomp/ner/LbjTagger/LearningCurveMultiDataset.java
```diff
@@ -78,16 +82,18 @@ public static void getLearningCurve(int fixedNumIterations, String trainDataPath
...
+        getLearningCurve(train, test, fixedNumIterations, incremental);
     }
 
     /**
-     * use fixedNumIterations=-1 if you want to use the automatic convergence criterion
+     * use fixedNumIterations=-1 if you want to use the automatic convergence criterion, incremental
+     * true will start with the existing models weights, and continue training with that set of default
+     * weights.
      * <p>
      * NB: assuming column format
```
cogcomp-dev: please use standard javadoc param descriptions

ner/src/main/java/edu/illinois/cs/cogcomp/ner/NerTagger.java
```diff
@@ -38,13 +38,13 @@ public static void main(String[] args) {
             Parameters.readConfigAndLoadExternalData(args[args.length - 1], areWeTraining);
 
             if (args[0].equalsIgnoreCase("-train"))
-                LearningCurveMultiDataset.getLearningCurve(-1, args[1], args[2]);
+                LearningCurveMultiDataset.getLearningCurve(-1, args[1], args[2], false);
             else if (args[0].equalsIgnoreCase("-trainFixedIterations"))
                 LearningCurveMultiDataset.getLearningCurve(Integer.parseInt(args[1]), args[2],
-                        args[3]);
+                        args[3], false);
             else {
```
cogcomp-dev: is it always the case that if not 'train' or 'trainFixedIterations', we definitely want incremental training?

cogcomp-dev: no, but in the "else" clause it is set to 'true' by default -- right?  That is the case I am asking about.

cowchipkid: There are no calls to getLearningCurve in the else clause, perhaps you are thinking of the "train" argument to ModelLoader.load, which is something else altogether. Although, I may want to change that, so if the model can not be found, it will throw. I will make that change and check it in.

cowchipkid: I have checked that change in.

cogcomp-dev: You're right -- thanks for pointing that out.

PR #675, opened 2018-08-02
bugs with multiple NERAnnotators per process space

cowchipkid:
There were some bugs with NER model loading, and all annotators were sharing the same NER model.

### DIFF
--- a/big-data-utils/pom.xml
+++ b/big-data-utils/pom.xml
@@ -3,7 +3,7 @@
     <parent>
         <artifactId>illinois-cogcomp-nlp</artifactId>
         <groupId>edu.illinois.cs.cogcomp</groupId>
-        <version>4.0.11</version>
+        <version>4.0.12</version>
     </parent>
 
     <modelVersion>4.0.0</modelVersion>
@@ -23,7 +23,7 @@
         <dependency>
             <groupId>edu.illinois.cs.cogcomp</groupId>
             <artifactId>illinois-core-utilities</artifactId>
-            <version>4.0.11</version>
+            <version>4.0.12</version>
         </dependency>
         <dependency>
             <groupId>org.xeustechnologies.google-api</groupId>
--- a/chunker/pom.xml
+++ b/chunker/pom.xml
@@ -2,7 +2,7 @@
     <parent>
         <artifactId>illinois-cogcomp-nlp</artifactId>
         <groupId>edu.illinois.cs.cogcomp</groupId>
-        <version>4.0.11</version>
+        <version>4.0.12</version>
     </parent>
 
     <modelVersion>4.0.0</modelVersion>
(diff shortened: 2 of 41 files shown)


nitishgupta:
Build is failing, btw. @cowchipkid

cowchipkid:
Ya, I am seeing that as well on my machine. Something broken on merge. I’ll update soon

nitishgupta:
@cowchipkid: looks like there was some orthogonal update to NER which now causes two (simple) merge conflicts. Could you please have a look? 

@mssammon: Can we merge this after Tom fixes the merge request?

mssammon:
@nitishgupta yes, just waiting for conflict resolution.

cowchipkid:
@danyaljj This is ready to merge, can you put your OK on it and merge?

cowchipkid:
@danyaljj @mssammon I've updated to version 4.0.12, fixed runNER so when tests complete, we are good.

mssammon:
Can merge when CI build finishes/shows green.

nitishgupta:
@cowchipkid I'm assuming you've tested multiple Annotators with different models running simultaneously. I had a question regarding the shared config you mentioned yesterday. What does that entail? How do we change the config programmatically when initializing the Annotators? What the configs we can change, and what are their possible values? Sorry, if this information is listed somewhere, but a quick read through the README didn't give me much info.

cowchipkid:
@nitishgupta Yes, different annotators will not use different models stored as fields of the NERAnnotator as they did before. However, other configuration properties like the one that enables the use of the level 2 model, the one that breaks sentences on newlines and so on are all shared. And there is nothing you can do about it. Wish I had a better answer for you, if it were easy, I would just make the config class factory provided keyed on view name. But that's not a one day job, as some of the low level feature generators use these configuration properties and feature generators do not have access to the view name.

danyaljj:
Looks good to me. Merge when you feel happy about it.

nitishgupta:
@cowchipkid Thanks for your response.! I think I understand your  Is there still a way to find out what the modifiable config parameters are, their possible values, and how to change them programmatically?

cowchipkid:
@nitishgupta All these properties are defined in a [ParametersForLbjCode](https://github.com/CogComp/cogcomp-nlp/blob/master/ner/src/main/java/edu/illinois/cs/cogcomp/ner/LbjTagger/ParametersForLbjCode.java), all these properties are static fields of this class, and a static embedded properties class. From where I sit, coming up with a good solution that works consistently is the most expedient way to fix this, it is impossible for multiple annotators to share static variables without tons of monitor locks, defeating the usefulness of multi-threading. I believe a much better solution is to completely re-architect this properties container, rather than a singleton, build a factory interface to serve up instances per annotator. The only quandary is then how the low level feature generators can share a key with the annotator to fetch the instance of the properties from the factory. If I had more time, I could fix this, but alas my last day...

nitishgupta:
Okay, got it. Thank you.

Issue #492, opened 2017-06-28
Sentence splitter mistake


danyaljj:
For this input: 
```
On each chromosome pair are smaller bodies (genes) which control all aspects of a bird, its colour, shape, bone structure, length of feather, size etc. The feather is a structure which is unique to birds.interpreted them as two paired rows of structures that are anatomically very much like feathers, and which are in positions like those of birds.
```
I expect two sentences, while it gives me only one sentence: 
http://austen.cs.illinois.edu:8080/annotate?text=On%20each%20chromosome%20pair%20are%20smaller%20bodies%20(genes)%20which%20control%20all%20aspects%20of%20a%20bird,%20its%20colour,%20shape,%20bone%20structure,%20length%20of%20feather,%20size%20etc.%20The%20feather%20is%20a%20structure%20which%20is%20unique%20to%20birds.interpreted%20them%20as%20two%20paired%20rows%20of%20structures%20that%20are%20anatomically%20very%20much%20like%20feathers,%20and%20which%20are%20in%20positions%20like%20those%20of%20birds.&views=

danyaljj:
It mistakenly splits on "Lord.": 

<img width="852" alt="screen shot 2017-10-21 at 7 55 38 pm" src="https://user-images.githubusercontent.com/2441454/31856826-f31997ea-b699-11e7-8e33-e500804e6128.png">

mssammon:
But that's a period after the first "Lord" ? And that is the first word in the text? If so, what other behavior would you expect?

danyaljj:
The previous example: agreed wiith Mark.

This example uses `;` as sentence splitter. Wondering if we can do anything about this: 
```
He increased the faculty by more than 500 professors ; the academic quality of the student body has improved dramatically , with the average SAT score rising from 1240 to 1360 ; the number of minority students more than doubled ; the endowment grew from $350 million to more than $3 billion ; the annual operating budget rose from $177 million to more than $650 million ; and annual research funding improved from $15 million to more than $70 million
```

danyaljj:
These are some questions from the SQuAD dataset and are supposed to be one sentence but are split into two sentences. 
I understand that some of them not are "standard", but after all any "human" would consider them one sentence and our sentence-splitter should have the same behavior as well. 

```
Who grouped 23 unpublished pieces and published them as Opp. 66-73 in 1855?
Who did a recording where Étude Op. 10 No. 5. follows relude Op. 28 No. 7?
[...]

PR #681, opened 2018-08-17
A temporary fix to sentence splitter behavior in TokenizerStateMachine when dealing with acronyms; Also fixing tokenizer unit tests to ensure test passing when using windows style line separator

### LINKED ISSUE #657: StatefulTokenizer error: lower case acronym ending with dot
The sentence "I was born in Urbana, IL. in 1992." is correctly recognized, but "I was born in Urbana, Il. in 1992." (IL with L in lower case) is split into two sentences.

schen149:
Adding a rule in [TokenizerStateMachine](https://github.com/CogComp/cogcomp-nlp/blob/master/tokenizer/src/main/java/edu/illinois/cs/cogcomp/nlp/tokenizer/TokenizerStateMachine.java) -- When the current character is a period; the next char is white space; and the next next char is a lower-cased alphabet, (I think) we can safely say that the next word would not be the start of a sentence, so the TSM should not split on the current period.  

This fix resolves #657, but doesn't work on the last example Daniel mentioned in #492. 

> " A Science Hall was built in 1883 under the direction of Fr. Zahm, but in 1950 it was converted to a student union building and named LaFortune Center, after Joseph LaFortune, an oil executive from Tulsa, Oklahoma."

Also fixed tokenizer unit tests to ensure test passing when using windows style line separator (\r\n).

### DIFF
--- a/tokenizer/src/main/java/edu/illinois/cs/cogcomp/nlp/tokenizer/TokenizerStateMachine.java
+++ b/tokenizer/src/main/java/edu/illinois/cs/cogcomp/nlp/tokenizer/TokenizerStateMachine.java
@@ -336,7 +336,7 @@ public class TokenizerStateMachine {
                                     if (getCurrent().isAbbr())
                                         return; // previous was upper case, acronym and word
                                                 // continues
-                                    else if (Character.isWhitespace(c) && Character.isLowerCase(nextnextChar))
+                                    else if (Character.isLowerCase(nextnextChar))
                                         return; // when the next char is white space and the next next char
                                                 // is lowercase, we know that the next word is not start of
                                                 // a sentence, so we continue.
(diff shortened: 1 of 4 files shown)


schen149:
@cowchipkid What do you think about this? 

For the "Fr. Zahm" example, I don't think there's a neat fix except adding "Fr" to the known acronym list.

Slash0BZ:
@schen149 I think adding Fr. to the known acronym list is a good way to go. Maybe you can add it here as well?

tokenizer/src/main/java/edu/illinois/cs/cogcomp/nlp/tokenizer/TokenizerStateMachine.java
```diff
@@ -332,9 +332,14 @@ public void process(char token) {
                                 } else {
                                     // check for all uppercase and periods back to the start of the
                                     // word or a "-"
+                                    char nextnextChar = peek(2);
                                     if (getCurrent().isAbbr())
                                         return; // previous was upper case, acronym and word
                                                 // continues
+                                    else if (Character.isWhitespace(c) && Character.isLowerCase(nextnextChar))
```
Slash0BZ: I think in this "else" clause, Character.isWhitespace(c) is redundant. Checking nnc to be lowercase looks good to me

cowchipkid:
If "Fr." is a lagitimate acronym, and it is, altho rarely used,  it should definitely be added to the acronyms list, that's the right way to go.

danyaljj:
What is this PR waiting for @schen149 ?

schen149:
Review from @cogcomp-dev ? Or do you want me to run a POS benchmark? @danyaljj

PR #687, opened 2018-08-30
Issue 665

### LINKED ISSUE #665: SpanLabelView may not enforce non-overlap constraint
It looks like if you call `addConstituent()` it sidesteps the span check:
https://github.com/CogComp/cogcomp-nlp/blob/master/core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/datastructures/textannotation/SpanLabelView.java#L62
This should perform a bounds check unless there is a good reason not to; otherwise, documentation should reflect the intention.

ChaseDuncan:
- Changed SpanLabelView.addConstituent(Constituent) to enforce nonoverlapping
  constituents when appropriate, i.e. the flag is set.
- Added a class with two basic tests to make sure that the changes work
  correctly.

Closes #665

### DIFF
--- a/core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/datastructures/textannotation/TokenLabelView.java
+++ b/core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/datastructures/textannotation/TokenLabelView.java
@@ -31,8 +31,17 @@ public class TokenLabelView extends SpanLabelView {
         this(viewName, viewName + "-annotator", text, 1.0);
     }
 
+    public TokenLabelView(String viewName, TextAnnotation text, boolean allowOverlappingSpans) {
+        this(viewName, viewName + "-annotator", text, 1.0, allowOverlappingSpans);
+    }
+
     public TokenLabelView(String viewName, String viewGenerator, TextAnnotation text, double score) {
         super(viewName, viewGenerator, text, score);
+    }
+
+    public TokenLabelView(String viewName, String viewGenerator, TextAnnotation text, double score,
+                          boolean allowOverlappingSpans) {
+        super(viewName, viewGenerator, text, score, allowOverlappingSpans);
     }
 
     /**
--- a/external/stanford_3.3.1/src/main/java/edu/illinois/cs/cogcomp/pipeline/handlers/StanfordTrueCaseHandler.java
+++ b/external/stanford_3.3.1/src/main/java/edu/illinois/cs/cogcomp/pipeline/handlers/StanfordTrueCaseHandler.java
@@ -54,7 +54,7 @@ public class StanfordTrueCaseHandler extends Annotator {
     public void addView(TextAnnotation ta) throws AnnotatorException {
         Annotation document = new Annotation(ta.text);
         pipeline.annotate(document);
-        TokenLabelView vu = new TokenLabelView(viewName, ta);
+        TokenLabelView vu = new TokenLabelView(viewName, ta, true);
 
         for (CoreMap sentence : document.get(CoreAnnotations.SentencesAnnotation.class)) {
             for (CoreLabel token : sentence.get(CoreAnnotations.TokensAnnotation.class)) {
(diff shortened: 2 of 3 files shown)


danyaljj:
Closes #665

core-utilities/src/main/java/edu/illinois/cs/cogcomp/core/datastructures/textannotation/SpanLabelView.java
```diff
@@ -60,6 +58,13 @@ public SpanLabelView(String viewName, String viewGenerator, TextAnnotation text,
 
     @Override
     public void addConstituent(Constituent constituent) {
+
```
mssammon: small improvement: please move start/end assignments inside if{} block, as they aren't being used otherwise.

core-utilities/src/test/java/edu/illinois/cs/cogcomp/annotation/SpanLabelViewTest.java
```diff
@@ -0,0 +1,100 @@
...
+import edu.illinois.cs.cogcomp.core.datastructures.textannotation.TextAnnotation;
+import edu.illinois.cs.cogcomp.nlp.tokenizer.Tokenizer;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class SpanLabelViewTest {
+    // test that addConstituent(Constituent) does not allow overlapping spans
```
mssammon: make the inline comment a class javadoc. comment can be exactly the same.

mssammon:
@mayhewsw , there were ways to add constituents that sidestepped the overlap check. This issue is intended to fix the problem.

mayhewsw:
Ah, I see it now. Yes, looks good.

cogcomp-dev:
@ChaseDuncan one test is failing in CI (teamcity build): please check into it and comment here if it is not related to the changes you made.

danyaljj:
TeamCity is failing: 
http://morgoth.cs.illinois.edu:5800/viewLog.html?buildId=762&buildTypeId=CogcompNlp_Build&tab=buildLog&filter=err

ChaseDuncan:
It looks like there are more places where this was being done unintentionally. Will look into it...

mssammon:
@ChaseDuncan looks like CI build is failing -- please take a look...

ChaseDuncan:
@mssammon I think the problem is a conflict in the design of mention detection and the some of the other CCG sofware like BasicAnnotatorService. 

For example:

By default MD appears to create overlapping spans, e.g.
[NAM-ORG the CDC ] [NOM-PER the CDC 's Dr. ] [NAM-PER the CDC 's Dr. Michael Jhung ] .

For some (probably good) reason I don't understand, there's some View copying happening in BasicAnnotatorService. By default it creates Views which do not allow overlapping spans.
So when it tries to copy the constituents from the MENTION View into a new View instance, it dies.
Previously this was done using addConstituent(Constituent) so the allowOverlappingSpans check was circumvented.

It's not immediately clear to me how best to address this since the problem involves a bunch of complex system for which I have basically no knowledge. But, if no one else knows what to do, I can look into the code and see what I can figure out.

mssammon:
I think it is reasonable to change any code that violates the assumption by changing it to explicitly allow overlapping constituents -- since that is going on anyway.  For posterity, you could create a list of any such places you change, and an issue that notifies everyone these were changed -- that would ensure (in theory) that *someone* checks each one.  If you want to make this your issue for the next k weeks, please go ahead and then ask anyone who seems likely to have a clue for each given component.  @Slash0BZ  and @danyaljj are likely candidates for who to ask w.r.t. the Mention view.

Slash0BZ:
Ah I didn't know that SpanLabelView by design discourages overlapped mentions. In MD, complete mention spans can overlap with each other, but mention heads cannot. I think we either need a special case for SpanLabelView that allows overlapped mentions, or I can try to only add mention heads to the view, and use separate function to generate the complete mention using information stored in the head.

mssammon:
@Slash0BZ I think the mention view should allow overlap, and that you should provide a utility method in the relevant code that verifies no constraints are violated.

PR #705, opened 2018-10-26
Email parsing fixes

cowchipkid:
Fixed some problems with email parsing where text spanning multiple lines might be captured as a single email address.


cowchipkid:
On my machine, running from at home, the tests fail on the Curator. I think this is because I am outside the U of I firewall, but I wouldn't swear to that.

tokenizer/src/main/java/edu/illinois/cs/cogcomp/nlp/tokenizer/TokenizerStateMachine.java
```diff
@@ -81,6 +81,8 @@
     
     /**
      * Init the state machine decision matrix and the text annotation.
+     * @param splitOnDash true to split tokens on a "-"
+     * @param splitOnTwoNewlines if true split a paragraph on two new lines.
```
danyaljj: Thanks!

danyaljj:
Looks good to me. Will merge when it turns green.

PR #711, opened 2018-12-31
JA tokenizer & stateful tokenizer fixes

cowchipkid:
Added a Japanese tokenizer to the multilingual tokenizer set, and fixed an issue with how periods are handled and some issues with email address where an email address later in a line captures when it should not.


danyaljj:
Thanks, Tom! Looks good to me. Please merge when you're feeling ready.

PR #712, opened 2019-01-17
ojAlgo v47

apete:
Upgrade ojAlgo from v40 to v47.

Info: ojAlgo has 3:d party iterations with CPLEX, GUROBI and MOSEK. If you can detect that any of those solvers are available then all you need to to is `model.addIntegation(INTEGRATION)` to register that integration with ojAlgo, and then that solver will be used. No code changes required.


danyaljj:
Thanks!

PR #718, opened 2019-05-03
Bug Fixes, Added methods to NERAnnotator

cowchipkid:
Added methods to the NERAnnotator to return a map of features, and the associate weights per model. And fixed a bug in the StatefullTokenizerTest. Also fixed a but in ModelLoader that would in some cases prevent finding a model in a jar file.


danyaljj:
Looks good to me. 

Currently, we don't have an automated test, so don't really whether all tests pass or not. But I trust you that you have checked it. 

Feel feer to merge when you're happy.

PR #720, opened 2019-06-15
Improve NerBenchmark flexibility

cowchipkid:
…d that can be passed

Refactored NerBenchmark to support builder pattern, improved flexibility. Obviously, NerBenchmark is no longer a static singleton.


danyaljj:
Looks good to me. Merging.

